PR #3, opened 2016-02-03
first go at better multi editing support

splitbrain:
This is an idea how to implement multiedit support. The default is to simply repeat the single edit field for each existing value + 1 empty. The field name is made an array then.

Types can override this default behavior and implement their own editor. If they don't post an array, the spitValues() method will be called before saving.


splitbrain:
BTW. I noticed that SchemaData::getData() returns arrays for single value fields as well. I guess there's a bug somewhere.

PR #21, opened 2016-02-17
Add *~ and = comp and adjust to data functionality

micgro42:


types/AbstractBaseType.php
```diff
@@ -242,15 +242,22 @@ public function renderMultiValue($values, \Doku_Renderer $R, $mode) {
...
-        } else if($comp == '!~') {
-            $sql = "$column NOT LIKE ?";
-            $opt = array('%' . $value . '%');
-        } else {
-            $sql = "$column $comp ?";
-            $opt = array($value);
+        switch ($comp) {
+            case '*~':
+                $sql = "$column LIKE ?";
+                $opt = array('%' . $value . '%');
```
splitbrain: This should happen in ConfigParser. When the config parser finds a ``*~`` it simply wraps the value with ``*`` and passes ``~`` as $comp. This way you have one case less in the many Types to handle.

types/AbstractBaseType.php
```diff
@@ -242,15 +242,22 @@ public function renderMultiValue($values, \Doku_Renderer $R, $mode) {
...
+                break;
+            case '~':
+                $sql = "$column LIKE ?";
+                $opt = array($value);
+                break;
+            case '!~':
+                $sql = "$column NOT LIKE ?";
+                $opt = array('%' . $value . '%');
```
splitbrain: This doesn't seem right. It should be the same as ``~`` just with a ``NOT LIKE``.

PR #24, opened 2016-02-22
Use php5 query builder

micgro42:


action/entry.php
```diff
@@ -217,17 +217,17 @@ protected function validate(AbstractBaseType $type, $label, $data) {
...
         }
 
+        if(!blank($data)) {
+            try {
+                $type->validate($data);
+            } catch(ValidationException $e) {
+                msg($prefix . $e->getMessage(), -1);
+                $ok = false;
+            }
+        }
```
splitbrain: Why did you move this outside the else? Wouldn't multivalue types be validated twice now? A non-empty array is !blank.

micgro42: No, the ``if(is_array($data))`` has an ``return $ok`` at its end.

splitbrain: Ah! I missed that.

Issue #6, opened 2016-02-17
implement type dt


splitbrain:
Accepts a date in the form YYYY-MM-DD

Should use the jQuery UI data picker for entry

Output should be formatable through a config option. Defaulting to dformat()

stvoigt:
It would be good to use a placeholder (current date) as a default value. (The way we use it in templates.)

PR #25, opened 2016-02-22
Add the date type

micgro42:
SPR-323, GitHub Issue #6

### DIFF
--- a/conf/default.php
+++ b/conf/default.php
@@ -5,4 +5,4 @@
  * @author Andreas Gohr, Michael Große <dokuwiki@cosmocode.de>
  */
 
-//$conf['fixme']    = 'FIXME';
+$conf['dformat']    = 'Y/m/d';
--- a/conf/default.php
+++ b/conf/default.php
@@ -5,4 +5,4 @@
  * @author Andreas Gohr, Michael Große <dokuwiki@cosmocode.de>
  */
 
-$conf['dformat']    = 'Y/m/d';
+# $conf['dformat']    = 'Y/m/d';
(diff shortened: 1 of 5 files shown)


conf/default.php
```diff
@@ -5,4 +5,4 @@
  * @author Andreas Gohr, Michael Große <dokuwiki@cosmocode.de>
  */
 
-//$conf['fixme']    = 'FIXME';
+$conf['dformat']    = 'Y/m/d';
```
splitbrain: This config should be in the type's config similar to the format we have in the Integer type.

types/Date.php
```diff
@@ -0,0 +1,50 @@
+<?php
+namespace plugin\struct\types;
+
+use plugin\struct\meta\ValidationException;
+
+class Date extends AbstractBaseType {
+
+    protected $config = array(
+        'prefix' => '',
+        'postfix' => '',
```
splitbrain: I think we should decide if we always want to support prefix and postfix and if so, move it to AbstractBasetype. Otherwise we should get rid of it where it's not needed. Maybe discuss with Detlef and Stefan

types/Date.php
```diff
@@ -0,0 +1,50 @@
...
+     * @return string html
+     */
+    public function valueEditor($name, $value) {
+        $name = hsc($name);
+        $value = hsc($value);
+        $html = "<input class=\"struct_date\" name=\"$name\" value=\"$value\" />";
+        return "$html";
+    }
+
+    public function validate($value) {
```
splitbrain: missing doc block

types/Date.php
```diff
@@ -0,0 +1,50 @@
...
+        list($year, $month, $day) = explode('-',$value);
```
splitbrain: the explode should be limited to 3 values (third parameter)

micgro42:
When merged, this closes #6

PR #26, opened 2016-02-22
Add image type

### LINKED ISSUE #13: implement type img
Image should be a local media ID or external URL. Configuration needs to cover display in page and in aggregation (you probably want a smaller thumbnail in aggregation than in the page itself). We need width and optionally height. Add ligtbox class for when gallery plugin is installed.

micgro42:
ToDo:

- [x]  Validation?
- [x]  Tests with external Images
- [x]  Gallery Plugin Integration / Lightbox Class
- [x] Width / Height? Currently automatically detected
- [x] Multieditor

This pull request is requires the changes of splitbrain/dokuwiki#1479. It also makes use of the changes of splitbrain/dokuwiki#1472.

When merged, this pull request closes #13.


micgro42:
Gallery / Lightbox:
* Klasse,
* Direct Link?
* Data-attribut zum zusammenfassen

types/Img.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Output the stored data
+     *
+     * @param string|int $value the value stored in the database
+     * @param \Doku_Renderer $R the renderer currently used to render the data
+     * @param string $mode The mode the output is rendered in (eg. XHTML)
+     * @return bool true if $mode could be satisfied
+     */
+    public function renderValue($value, \Doku_Renderer $R, $mode) {
+        if (!empty($value)) {
```
splitbrain: I believe we do not render empty values? We should set up some rules on what types do not need to check themselves (like validate() is never called for blank fields).

syntax/table.php
```diff
@@ -320,10 +320,11 @@ protected function buildColumnHeaders($mode, Doku_Renderer $renderer, $data) {
     /**
      * @param string        $mode     the mode of the renderer
      * @param Doku_Renderer $renderer the renderer
+     * @param string        $hash     hash to identify the table and group images in gallery
      */
-    protected function startScope($mode, \Doku_Renderer $renderer) {
+    protected function startScope($mode, \Doku_Renderer $renderer, $hash) {
         if ($mode == 'xhtml') {
-            $renderer->doc .= '<div class="table structaggegation">';
+            $renderer->doc .= "<div class=\"table structaggregation hash:$hash\">";
```
splitbrain: I'm not sure if this is a valid class name. Why not ```data-hash=$hash```?

types/Img.php
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            $html = $R->externalmedia($value, null, null, null, null, null, 'direct', $returnLink);
+        }
+
+        if ($mode == 'xhtml') {
+            $hash = '';
+            if (strrpos($R->doc,'</table>') < strrpos($R->doc,'class="table structaggregation')) {
+                $hash = substr($R->doc,strpos($R->doc,'hash:',strrpos($R->doc,'class="table structaggregation'))+5,32);
+                $hash = "[gal-$hash]";
+            }
```
splitbrain: Ah, now I understand what you meant. There's an easier way. The renderer object you are passed here is the same object you used earlier in startScope. You could simply store the hash in $renderer->info['struct_table_hash'] in startScope and remove it again in stopScope. Later check here, if the value is set and act accordingly.

Issue #40, opened 2016-02-29
implement type user


stvoigt:
Users in data entries/forms/tables were a very nice feature in my past wikis. The only way to achieve this was a workaround with usercreate page and the page-data-type looking for existing pages in the user namespace. It would be nice to have kind of user field with autocompletion. Even better would be autocompletion on both: user name and real name. Or maybe something like the @-Link here in github editor...

stvoigt:
It would be good to allow "non-existing users" too. If we document a meeting with participants it would be great to add users from the wiki (to link to user-profile) and add other participants (that do not have to be users). Right now, you cannot save values that are non-existing users.

PR #41, opened 2016-02-29
WIP: Implement Type User

splitbrain:
This adds an auto completing User type as suggested in #40 

Searching in Full Names is optional and can be disabled in config. Filtering by Full Name is not and will not be supported.

- [x] validation
- [x] resolve to full name in display
- [x] test behavior with multi fields
- [x] fix tests (we really need to stop doing exact HTML checks)


splitbrain:
This branch also implements auto completion for the page type (reusing stuff I did for the user type)

PR #65, opened 2016-03-15
Add tests for a schema consisting only of multis

micgro42:
This test reproduces the current behaviour of the plugin. Is this the SQL we expect?


splitbrain:
A test checking for the correct data would be more helpful I guess. The SQL might change but we want to be sure the result is correct.

PR #104, opened 2016-06-14
Invalidate the cache when a schema changes

micgro42:
So Schema changes are visibly immediately.

SPR-347


splitbrain:
I think these touches should refer to the config cascade (writing to the last local file configured)

PR #113, opened 2016-07-11
Fix: Show single value in multi until overwritten

micgro42:
If a field was changed from single to multi value, then the former single value appeared to be "lost". Now it is still shown.

This pull-request depends on cosmocode/sqlite#38

ToDo:
- [ ] finish cosmocode/sqlite#38
- [x] Add this functionality to ``Search::execute()`` as well


micgro42:
Add this functionality to ``Search::execute()`` as well

micgro42:
Closed in favor of #160

PR #114, opened 2016-07-12
Do not show rows in aggregations if they're empty

micgro42:
There was no check to hide rows from empty schemas in aggregations.

SPR-541

### DIFF
--- a/meta/Search.php
+++ b/meta/Search.php
@@ -390,7 +390,7 @@ class Search {
         foreach($this->columns as $col) {
             $val = $rowColumns["C$C"];
             $C += 1;
-            if (empty($val) || is_a($col->getType(),'dokuwiki\plugin\struct\types\Page') && $val == $rowColumns["PID"]) {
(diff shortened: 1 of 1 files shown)


meta/Search.php
```diff
@@ -376,6 +379,25 @@ public function findColumn($colname) {
...
+     *
+     * @param array $rowColumns an array as returned from the database
+     * @return bool
+     */
+    private function isRowEmpty($rowColumns) {
+        $C = 0;
+        foreach($this->columns as $col) {
+            $val = $rowColumns["C$C"];
+            $C += 1;
+            if (empty($val) || is_a($col->getType(),'dokuwiki\plugin\struct\types\Page') && $val == $rowColumns["PID"]) {
```
splitbrain: A row containing a 0 might not be empty. I guess we should compare against ```=== ''```  here?

micgro42: Done

PR #126, opened 2016-07-13
WIP: use query builder for creating sql queries

splitbrain:
The idea is to give types much more control over the SQL that selects and searches data to allow for more complex types. For this they need to have a defined access to the query that selects the data. This branch adds a query builder for this...

early work in progress

@micgro42 help with testing (even tests for missing functionality) greatly appreciated

- [x] add GROUP BY support
- [ ] add LIMIT/OFFSET support
- [x] add SORT BY support
- [x] make Search use the builder
- [x] give Types the builder in compare()
- [x] make SchemaData use the builder
- [x] give Types the builder in a new getData() function
- [ ] escape/quote passed table and columns names


splitbrain:
Search now uses the QueryBuilder in the most simplistic way. Next the types should be made aware of the builder and we need to figure out what else they need to effectively use it

micgro42:
It's only a matter of personal preference, but could you use
```php
if (!isset($this->from[$tablealias])) {
    throw new StructException('Table Alias does not exist');
}
```
instead of
```php
if(!isset($this->from[$tablealias])) throw new StructException('Table Alias does not exist');
```
?

That way it will be easier to spot lines not covered by tests and it's less error prone when extending it with an additional debug/log/error/cleanup-statement

splitbrain:
fair enough. I'll adjust the code accordingly.

types/AbstractBaseType.php
```diff
@@ -385,6 +386,33 @@ public function compare($column, $comp, $value) {
...
+            case '~':
+                $QB->filters()->where($type, "$column LIKE $pl");
+                break;
+            case '!~':
+                $QB->filters()->where($type, "$column NOT LIKE $pl");
+                break;
+            default:
+                $QB->filters()->where($type, "$column $comp $pl");
+        }
+    }
```
splitbrain: @micgro42 this is a first idea how the QueryBuilder could be passed on to the type (compare with the above compare function for what we currently have). I'm not sure this is enough info, yet. I guess we would also need some tablenames/aliases (eg. the data and multi tables this type's column is part of).

Also this is currently not very flexible in terms of inheritance. A type would always need to implement the function on its own completely and couldn't inherit any functionality from this function (eg. call parent::filter()).

Any better ideas to do this?

micgro42: In the current master  the ``$column`` argument for ``compare()`` in ``Search::getSQL()`` is already constructed to consist of the right aliases/column references. This should be done for the ``filter()`` method as well. However I wonder more about the ``$value`` argument. What if that actually refers to a value in column of another table as well instead of a given fix value?

I think it would be helpful for me, if we can write down some concrete use cases which need to be covered, rather than talk about it in abstract terms. I see if I can get started with this today.

Some style questions:
- Could we change the ``$type`` parameter to ``$op`` or ``$operator``? It gets confusing because we already are in class named ``Type``.
- I'm not sure about this myself: But would it be a good idea to put the line ``$QB->filters()->where($type, "$column $comp $pl");`` after the switch and have the switch only contain lines like ``$comp = 'LIKE';``?

splitbrain: All very good suggestions. Can you implement the style changes?

splitbrain:
@micgro42 since 020b6ad39c9316189b42b1ecb587af0ef6fcb9c0 our tests are broken because I had to change something within SchemaData. Would be great if you could help with fixing.

splitbrain:
The page type now has a ``usetitles`` option which makes use of an additional titles table. It seems to work so far but requires tests. As always, help is appreciated ;-)

splitbrain:
regexp filtering is now included in this branch and also needs testing. I used ``=*`` to indicate a regex filter, not sure if that's a good symbol.

PR #127, opened 2016-07-14
[WIP] Add tests for QueryBuilder

micgro42:
Here more tests for the query builder branch will be added. The query builder branch will be regularly merged into this tests-branch to prevent divergence.

### DIFF
--- a/_test/QueryBuilderFrom.test.php
+++ b/_test/QueryBuilderFrom.test.php
@@ -38,4 +38,24 @@ class QueryBuilderFrom_struct_test extends StructTest {
         $qb->addTable('second', 'T1');
     }
 
+    /**
+     * @expectedException \dokuwiki\plugin\struct\meta\StructException
+     */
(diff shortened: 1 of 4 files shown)


_test/QueryBuilderFrom.test.php
```diff
@@ -0,0 +1,31 @@
...
+        $qb = new QueryBuilder();
+
+        $qb->addTable('first', 'T1');
+        $qb->addTable('second', 'T2');
+        $qb->addTable('third', 'T3');
+
+        $qb->addLeftJoin('T2', 'fourth', 'T4' , 'T2.foo=T4.foo');
+
+        $expectedSQL = '
+            SELECT FROM first AS T1, second AS T2, LEFT OUTER JOIN fourth AS T4
```
splitbrain: this is wrong SQL, there may not be a comma after T2. I just ran into this. Should be fixed in devel branch.

Issue #132, opened 2016-07-21
title: Have %title% column placeholder


splitbrain:
It makes sense to have a title equivalent to %pageid%

micgro42:
Dynamic filters for %title% columns result in a filter by %pageid%

micgro42:
Works now as expected.

Issue #131, opened 2016-07-21
title: $STRUCT.column$ contains raw JSON


splitbrain:
The column filter place holder does not work as expected for the new Page.usetitles=true

micgro42:
Scheint zu funktionieren

Issue #130, opened 2016-07-21
title: filter by pageid AND title


splitbrain:
Filters should match against title and pageid for Page.usetitles=true using an OR combination.

micgro42:
Seems to work as expected

Issue #129, opened 2016-07-21
title: restoring an old revision breaks title data


splitbrain:
When a revision of a page using the page.usetitles=true type is restored, the content of that column gets messed up because it tries to save the raw JSON.

micgro42:
Restoring works only when there is also a change in the text of page to be restored. Restores of struct-only changes simply do not work even so there is a success message.

Not sure that this is actually a title issue.

splitbrain:
Hmm I think I see the problem. Not title related.

micgro42:
Works as expected

PR #133, opened 2016-07-21
Title fixes

splitbrain:
Implements fixes for #129 #130 #131 and #132 

Needs tests and review @micgro42


splitbrain:
the two things you mentioned, should be fixed @micgro42 - can you have another look?

micgro42:
These four issues seem to be fixed now. However I think I have two new things:

- We don't need the ``$TITLE$`` placeholder anymore, now that ``%title%`` and title fields filter by the pageid as well. Right?
- I forgot the second, but I will remember it at some point.

Issue #58, opened 2016-03-08
Inline Editing for Aggregations


splitbrain:
It would be nice to have a way to edit all available data of a schema in one go through a big table. Implementing this would be half way to #37 

- [ ] Related to #59 a CSV import and export for this would be nice to have

PR #143, opened 2016-08-03
Inline editing

splitbrain:
Implements #58.

Contains parts of #142 (for easier manual testing), should merged after it.


splitbrain:
@micgro42 please have a look and suggest what should be tested.

micgro42:
Looks good to me.
This is only very optional and only implement it if it's simple to do:

But could do the check of whether opening the popup to the right or the left also for the autocompletion-popup?

micgro42:
Another thing I noticed: If images are used, then only the area of the cell around the image is clickable, i.e. the padding of the `<td>` element. That is ok, but it is possibly not obvious where the image ends and the padding starts. So could we give the images a black border or something similar?

micgro42:
Also I found a bug:
- [ ] The inline-edit for multi-image fields only always uses the first image-input

micgro42:
The issues mentioned above is fixed for cells that originally contain no media. However that bug still persists when the cell already contains one or more images when the page is loaded.
- [ ] multi-image when already at least one image exists uses always the first input

splitbrain:
I just noticed that this bug is present in "normal" edits too: #152

micgro42:
Works for me now :+1:

Issue #37, opened 2016-02-25
Have non-page tables


splitbrain:
As a future feature it might be nice to be able to create data tables with data not bound to pages. For these editing would happen directly on a page where users can add new, browse, edit or delete existing  entries.

Functionality like this has been asked for in the forum: https://forum.dokuwiki.org/thread/12820

It could also be used as a data source for the proposed dropdown type on #36

splitbrain:
# Concept

Some rough ideas on how to implement that

## Basic Principles

* we call those tables "lookup"
* they are not attached to any page
* the do not store data revisioned (there's only one active dataset) - however the schema for a lookup table would be revisioned
* it would probably be good to implement #53 because the page edit permission would no longer apply
* the inline editor from #143 provides the basis for editing the data of these tables.
  * edits would be done through a lookup table being embedded into a page somewhere
  * #120 has a suggestion of making this editor available from the admin backend, too

## Database Changes

* [ ] the tables for lookups would have the exact same structure as our current data_ and multi_ tables
  * the exception is that the ``pid`` column would be an auto increment integer instead of a string
  * ``rev`` would always be ``0`` and ``latest`` would always be 1 (no revisions, the data is always current)
* [ ] the ``schemas`` table would get an additiona ``islookup`` column

Keeping those tables so similar should make it possible to implement this new feature with minimal changes to existing classes

## Code Changes

* class ``Schema``
  * [ ] a new ``isLookup()`` method is added
* class ``SchemaBuilder``
  * [ ] the ``$data`` passed in the constructor needs to be extended with the ``islookup`` property
  * [ ] schemas can not be converted between lookup and non lookup!
* class ``Search``
  * [ ] Aggregations should not be able to select more than one Lookup
  * [ ] It should not be possible to select Non-Lookups mixed with Lookups (we can't join on ``pid``)
  * Types might join lookup tables via their ``select()`` and ``filter()`` methods
* Inline Editor (action, JS)
  * [ ] needs to differentiate between lookups and non-lookups
  * [ ] lookups do not create page revisions, create page locks
  * [ ] needs to be extended to add new rows (lookups only)
  * [ ] needs to be extended to delete rows (lookups only)
* [ ] new ``syntax/lookup``
  * basically limited version of the aggregation table syntax
  * always shows all columns
  * accepts a single lookup table only
* type ``Dropdown``
  * [ ] extend to be able to use a lookup as values
[...]

PR #147, opened 2016-08-09
WIP: Lookups aka Non-Page Tables

splitbrain:
implements #37 includes #143 

this is not done, yet. PR just for the sake of easier code review by @micgro42


splitbrain:
Current State: all the features mentioned on #37 are now implemented. Except for the Type support (Dropdown, Lookup). Before that is tackled, we should probably check which features need some more tests first.

splitbrain:
- [ ] CSRF tokens are not checked on row deletion
- [ ] new rows need to be validated

splitbrain:
Looking into the validation, I think we need to do another refactoring. The Validator class makes no sense as it is today. It has multiple problems:

* it instantiates new AccessTables (which creates overhead)
* it currently focuses entirely on page structs

I think the actual validation and check if data has changed mechanism needs to be moved into the AccessTable classes. The Validator could then maybe wrap around it and provide the initialized AccessTable objects for saving....

splitbrain:
* [ ] currently completely empty rows can be added

PR #149, opened 2016-08-09
Better Time mechanisms for AccessTable

splitbrain:
@micgro42 this should implement what we talked about. Can you have a look?

### DIFF
--- a/meta/AccessTable.php
+++ b/meta/AccessTable.php
@@ -152,7 +152,6 @@ abstract class AccessTable {
      * @return Value[] a list of values saved for the current page
      */
     public function getData() {
-        $this->getLastRevisionTimestamp($this->pid, $this->ts);
         $data = $this->getDataFromDB();
         $data = $this->consolidateData($data, false);
         return $data;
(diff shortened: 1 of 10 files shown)


meta/AccessTable.php
```diff
@@ -150,7 +152,7 @@ public function getDataColumn($column) {
      * @return Value[] a list of values saved for the current page
      */
     public function getData() {
-        $this->setCorrectTimestamp($this->pid, $this->ts);
+        $this->getLastRevisionTimestamp($this->pid, $this->ts);
```
micgro42: `getLastRevisionTimestamp()` now only returns a timestamp, without sideeffects. Shouldn't we use the retrieved value somehow?

meta/AccessTable.php
```diff
@@ -164,7 +166,7 @@ public function getData() {
      * @return array
      */
     public function getDataArray() {
-        $this->setCorrectTimestamp($this->pid, $this->ts);
+        $this->getLastRevisionTimestamp($this->pid, $this->ts);
```
micgro42: Same as [Line 155](https://github.com/cosmocode/dokuwiki-plugin-struct/pull/149/files#diff-48fa08b6d4c40117991ecc86ee1f24e0R155).

micgro42:
Except the one thing I annotated it looks good. :+1:

Issue #148, opened 2016-08-09
Empty field of page type with title enabled is not treated as empty


micgro42:
This might be related to #114, but also affects the data shown on the page itself.

splitbrain:
I'm not sure I understand what exactly the bug is here. Don't we store empty strings for all Types when a field is empty?

micgro42:
No, we store `null` for all types. Except for the page type, apparently.

micgro42:
Some screenshots that illustrate the problem:
![foopattern](https://cloud.githubusercontent.com/assets/7372507/17553612/08814024-5f07-11e6-8c39-b871d9dfe539.png)
![aggrbar](https://cloud.githubusercontent.com/assets/7372507/17553614/08a50a68-5f07-11e6-9d7c-77e0bf11d10e.png)
![rqpattern](https://cloud.githubusercontent.com/assets/7372507/17553613/08a1913a-5f07-11e6-93d0-9423f1dfdbad.png)
However the error only appears when the title-option is used. I'll adjust this issues title accordingly.

splitbrain:
Problem was not that JSON was returned. The problem was that isEmptyRow did not check against the raw value.

splitbrain:
Okay this a bit more complicated than I thought. I think we need to fix our rawValue concept which is currently all over the place. I'd like to make it like this:

* Change the ``Value`` class
  * [ ] to have a ``getRawValue`` method
  * [ ] to have a ``isEmpty`` method
* Change the ``DataSchema`` aka ``AccessTable`` class and
  *  [ ] remove the ``optionRawValue`` mechanism
  *  [ ] instead ``getData()`` returns ``Value[]``
  *  [ ] and ``getDataArray()`` always returns an array of raw values
* [ ] Change all code occurrences to make use of this new system

Because this affects the AccessTable class, I would like to either
  * make these changes in the lookup branch only
  * merge the lookup branch first

PR #153, opened 2016-08-11
Empty Handling

splitbrain:
Handling empty and raw values better as outlined in #148

### DIFF
--- a/action/diff.php
+++ b/action/diff.php
@@ -50,7 +50,6 @@ class action_plugin_struct_diff extends DokuWiki_Action_Plugin {
         $event->result .= "\n---- struct data ----\n";
         foreach($tables as $table) {
             $schemadata = AccessTable::byTableName($table, $id, $rev);
-            $schemadata->optionRawValue(true);
             $event->result .= $schemadata->getDataPseudoSyntax();
         }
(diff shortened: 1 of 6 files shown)


splitbrain:
It seems this is all that is to be done. Tests are all still passing and the problems outlined in #148 seem to be gone. @micgro42 can you have a second look?

micgro42:
the code looks good, hover I can not manually test it, since I get an error:
```
HY000 1 no such column: islookup:
SELECT DISTINCT tbl FROM schemas WHERE islookup = 1 ORDER BY tbl
```

splitbrain:
the migration probably didn't apply for you because of your own changes in
#154

On Aug 11, 2016 17:42, "Michael Große" <notifications@github.com> wrote:

the code looks good, hover I can not manually test it, since I get an error:

HY000 1 no such column: islookup:
SELECT DISTINCT tbl FROM schemas WHERE islookup = 1 ORDER BY tbl

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/cosmocode/dokuwiki-plugin-struct/pull/153#issuecomment-239201018>,
or mute the thread

PR #154, opened 2016-08-11
Add helper-API for export/import

micgro42:
This enables cosmocode/dokuwiki-plugin-farmsync#5


helper/imexport.php
```diff
@@ -0,0 +1,71 @@
...
+            $pattern = $this->sqlite->escape_string($pattern);
+            $sql[] = "INSERT INTO schema_assignments_patterns (pattern, tbl) VALUES ('$pattern','$schema')";
+        }
+
+        $this->sqlite->doTransaction($sql);
+        $assignments = new \dokuwiki\plugin\struct\meta\Assignments();
+        $assignments->propagatePageAssignments($schema);
+    }
+
+    public function getSchemaAssignmentPatterns($schema) {
```
splitbrain: missing doc blocks

meta/SchemaImporter.php
```diff
@@ -72,7 +72,7 @@ public function __construct($table, $json) {
                 $data['new'][] = $column;
             }
         }
-
+        $this->user = $input['user'];
```
splitbrain: I guess for your use case it makes sense to be able to set the user via the import data. However for manual imports I would expect that the user running the import is saved here rather than whatever is in the input. Otherwise it's impossible to see who messed up a schema by an import in the planned history feature.

_test/json/schema1.struct.json
```diff
@@ -2,6 +2,7 @@
     "structversion": "2016-03-21",
     "schema": "schema1",
     "id": "1",
+    "user": "",
```
splitbrain: Should this be there? I'd think we shouldn't export nor import the user into/from the json files.

helper/imexport.php
```diff
@@ -0,0 +1,103 @@
...
+     * Delete all existing assignment patterns of a schema and replace them with the provided ones.
+     *
+     * @param string   $schemaName
+     * @param string[] $patterns
+     */
+    public function replaceSchemaAssignmentPatterns($schemaName, $patterns) {
+        /** @var \helper_plugin_struct_db $helper */
+        $helper = plugin_load('helper', 'struct_db');
+        $this->sqlite = $helper->getDB();
+        $schemaName = $this->sqlite->escape_string($schemaName);
```
splitbrain: I'm not a big fan of escaping values ourselves. Though I understand why you do it. I think that's the reason I usually don't use the ``doTransaction()``  method.

helper/imexport.php
```diff
@@ -0,0 +1,103 @@
...
+        $sql = 'SELECT pattern FROM schema_assignments_patterns WHERE tbl = ?';
+        $res = $this->sqlite->query($sql, $schemaName);
+        $patterns = $this->sqlite->res2arr($res);
+        $this->sqlite->res_close($res);
+        return array_map(function($elem){return $elem['pattern'];},$patterns);
+    }
+
+    /**
+     * Get the json of the current version of the given schema.
+     * If the schema doesn't exist, then the returned JSON will have id 0.
```
splitbrain: Wouldn't it make more sense to throw an exception then?

helper/imexport.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public function importSchema($schemaName, $schemaJSON, $user = null) {
+        try {
+            $importer = new \dokuwiki\plugin\struct\meta\SchemaImporter($schemaName, $schemaJSON);
+            if (!blank($user)) {
+                $importer->setUser($user);
+            }
+            $ok = $importer->build();
+        } catch(dokuwiki\plugin\struct\meta\StructException $e) {
+            msg(hsc($e->getMessage()), -1);
```
splitbrain: since this is a helper method, I would prefer leaving the error handling to the caller, eg. just throw the exception.

meta/Schema.php
```diff
@@ -281,6 +292,7 @@ public function toJSON() {
             'structversion' => $this->structversion,
             'schema' => $this->getTable(),
             'id' => $this->getId(),
+            'user' => $this->getUser(),
```
splitbrain: again, I wouldn't export the user

meta/SchemaBuilder.php
```diff
@@ -140,8 +142,8 @@ protected function fixLabel($wantedlabel, &$labels) {
     protected function newSchema() {
         if(!$this->time) $this->time = time();
 
-        $sql = "INSERT INTO schemas (tbl, ts) VALUES (?, ?)";
-        $this->sqlite->query($sql, $this->table, $this->time);
+        $sql = "INSERT INTO schemas (tbl, ts, user) VALUES (?, ?, ?)";
+        $this->sqlite->query($sql, $this->table, $this->time, blank($this->user) ? $_SERVER['REMOTE_USER'] : $this->user);
```
splitbrain: I would initialize $this->user in the constructor, then you won't need this ugly construct.

helper/imexport.php
micgro42: I think it would make sense to extend the `doTransaction()` method of the sqlite-plugin to be able to handle sqlite-statements with extra value-arrays as well.

helper/imexport.php
micgro42: Not sure, because a missing schema is in this usecase not an error per-se. However if we'd introduce a `schemaExists($schemaName)` function, then we could treat a non-existing schema here as an error.

Issue #155, opened 2016-08-12
Inline editing corrupts page fields with title option


micgro42:
When a page has a page field with the use-title option enabled and the data of another field is edited via inline editing, then the data of the page field is corrupted.

Introduced in #143

PR #156, opened 2016-08-12
Fix broken page fields with title-option

micgro42:
helper_plugin_struct::saveData() expects raw values

I am not entirely sure, that this is the right place to fix this.

See #155 :beetle:


splitbrain:
I think this might be fixed in #153 already.

PR #158, opened 2016-08-15
Use Lookups in Dropdowns

splitbrain:
Another feature for #147 

This basically introduces the linking to a different page type, which can be quite complex.

review welcome

### DIFF
--- a/types/AbstractBaseType.php
+++ b/types/AbstractBaseType.php
@@ -451,6 +451,17 @@ abstract class AbstractBaseType {
     }
 
     /**
+     * This is called when a single string is needed to represent this Type's current
+     * value as a single (non-HTML) string. Eg. in a dropdown or in autocompletion.
+     *
+     * @param string $value
+     * @return string
+     */
+    public function displayValue($value) {
+        return $this->rawValue($value);
+    }
+
+    /**
      * Validate and optionally clean a single value
      *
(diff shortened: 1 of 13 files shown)


types/AbstractBaseType.php
```diff
@@ -451,6 +451,17 @@ public function rawValue($value) {
     }
 
     /**
+     * This is called when a single string is needed to represent this Type's current
+     * value as a single (non-HTML) string. Eg. in a dropdown or in autocompletion.
+     *
+     * @param string $value
+     * @return string
+     */
+    public function displayValue($value) {
```
micgro42: I'm curious about the naming. Whenever there is a difference between `rawValue` and "non-raw-value", isn't the raw value the one we do _not_ want to display?

splitbrain: I'm not sure I understand your issue.

* value - whatever was *selected* from the database (eg. a complex JSON string)
* rawValue - whatever is *stored* in the database column in the data_* table
* displayValue - a single, non-HTML string derived from the value

For most fields all three values are exactly the same. For others they are not. Eg. Page with Titles.

micgro42: Mh, I think `stringifyValue()` or `getValueString()` might be more speaking names for that function?

Issue #140, opened 2016-07-28
data in multi page fields lost, when doing preview


micgro42:
When a struct schema contains a field with page type, the content of that field is lost when doing a preview during edit.

micgro42:
This Bugs still exists in #158

micgro42:
So, I looked into this more specific. The Problem is that the $INPUT-data ([action/entry.php#L215](https://github.com/cosmocode/dokuwiki-plugin-struct/blob/72b4cd73ad0f3b8abb45dd442742f49fe03dd3e6/action/entry.php#L215) ) used to overwrite the data from the database contains only raw Values. However the setValue function [called in action/entry.php#L230](https://github.com/cosmocode/dokuwiki-plugin-struct/blob/72b4cd73ad0f3b8abb45dd442742f49fe03dd3e6/action/entry.php#L230) expects a complete value, i.e. in the case of a page-type with title option a JSON of pageid and title. Hence the setValue function sets that value to an empty array.
If we fix the setValue function to simply set the rawvalue for all (value, rawvalue, display) value variables, then the makeField() function [called in action/entry.php#L232](https://github.com/cosmocode/dokuwiki-plugin-struct/blob/72b4cd73ad0f3b8abb45dd442742f49fe03dd3e6/action/entry.php#L232) breaks our field because it uses only the Value::value variable and calls rawValue on that one again instead of directly using the Value::rawValue.

micgro42:
Unfortunately this issue is now fixed. See [Requirement:Brandschutz](https://demo.sprintdoc.de/projects:bau_des_flughafens_ber:requirements:terminal_brandschutz)

PR #159, opened 2016-08-19
Fix pages in multi-page-title-fields being lost during preview

micgro42:
This may not be the perfect place to fix it, but every alternative would involve possibly invasive design changes.
For more details and problem description see issue #140.

### DIFF
--- a/types/Page.php
+++ b/types/Page.php
@@ -162,6 +162,10 @@ class Page extends AbstractMultiBaseType {
      */
     public function rawValue($value) {
         if($this->config['usetitles']) {
+            if (cleanID($value) == $value) {
+                // This is already a raw value
+                return $value;
+            }
             list($value) = json_decode($value);
         }
         return $value;


types/Page.php
```diff
@@ -162,6 +162,10 @@ public function sort(QueryBuilder $QB, $tablealias, $colname, $order) {
      */
     public function rawValue($value) {
         if($this->config['usetitles']) {
+            if (cleanID($value) == $value) {
+                // This is already a raw value
+                return $value;
+            }
```
splitbrain: I don't like this. I think we should go the hard way and fix this properly in entry.php (or where ever).

I wish strings where real objects and we could enhance them with a flag to note what kind of value it is. Hmm... maybe we should do that? *sigh* those complex types introduced so much hardship, I wonder if they were worth it...

splitbrain:
See comment here https://github.com/cosmocode/dokuwiki-plugin-struct/pull/165/files#diff-bd53b24644eadbf3ca54bf9fd803bc7fR80

splitbrain:
Not as simple as I thought. I think the proper way is to give types a rawToComplex method...

splitbrain:
replaced by #166

PR #160, opened 2016-08-19
Write last value to multi-table on col change single->multi

micgro42:
The Problem we are trying solve is that the current value "disappears" when a column is made into a multi-column. The reason for this is that there is not yet a value for the current revision in the multi-table and we do not look into the data-table anymore for that column. Hence on change from single to multi we write the current value of that column for every page into the multi table.

Review welcome.

This is intended to replace and therefore closes #113.

### DIFF
--- a/meta/SchemaBuilder.php
+++ b/meta/SchemaBuilder.php
@@ -180,6 +180,9 @@ class SchemaBuilder {
                     if(!$res) return false;
                     $newTid = $this->sqlite->res2single($res);
                     $this->sqlite->res_close($res);
+                    if ($oldEntry['ismulti'] == false && $newEntry['ismulti'] == '1') {
+                        $this->migrateSingleToMulti($this->oldschema->getTable(), $column->getColref());
+                    }
                 }
             } else {
                 $enabled = false; // no longer there for some reason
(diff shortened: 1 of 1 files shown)


meta/SchemaBuilder.php
```diff
@@ -199,6 +202,26 @@ protected function updateColumns() {
         return true;
     }
 
+    private function migrateSingleToMulti($table, $colref) {
```
splitbrain: should be ``protected``. is missing a doc block

meta/SchemaBuilder.php
```diff
@@ -199,6 +202,26 @@ protected function updateColumns() {
...
+        $sqlSelect = "SELECT pid, rev, col$colref AS value FROM data_$table WHERE latest = 1";
+        $res = $this->sqlite->query($sqlSelect);
+        $valueSet = $this->sqlite->res2arr($res);
+        $this->sqlite->res_close($res);
+        $valueString = array();
+        foreach ($valueSet as $values) {
```
splitbrain: Under what circumstances can the select above return more than one row? There should be only one ``latest`` entry for one column in a data_* table!?

micgro42: There is one `latest` entry per page. We need to migrate the data for every page.

splitbrain: Ah okay. I think I got confused by the join later down which I thought is joining multiple values into a single column. But it's not. See my comment there then ;-)

meta/SchemaBuilder.php
```diff
@@ -199,6 +202,26 @@ protected function updateColumns() {
...
+            if (blank($values['value']) || trim($values['value']) == '') {
+                continue;
+            }
+            $valueString[] = "($colref, '$values[pid]', $values[rev], 1, '$values[value]')";
+        }
+        if (empty($valueString)) {
+            return;
+        }
+        $valueString = join(',', $valueString);
+        $sqlInsert = "INSERT OR REPLACE INTO multi_$table (colref, pid, rev, row, value) VALUES $valueString";
```
splitbrain: You should not build your own queries like this. In fact your actual value is not correctly escaped now and introduces a SQL injection problem.

micgro42: Yes, I incorrectly assumed that since I handle DB-values, not direct user input, that escaping would be unnecessary. But that turned out to be wrong.

This should be fixed with the latest commit.

PR #161, opened 2016-08-19
Do not hide rows if table is only pid and/or rev

### LINKED ISSUE #150: selecting %pageid% only doesn't work
For some reason I don't fully understand yet it's not possible to select only `%pageid%`. The search query returns an empty result set. As soon as you add another column it works.

micgro42:
This hides rows in aggregation-tables/results if they are empty except for the `%pageid%` or `%lastupdated%` column. However if a table consists solely of a `%pageid%` and/or a `%lastupdated%` column, then no rows will be skipped.

However the second part has the consequence that it is not possible to remove a page from such a table by deleting all its struct data for that schema. The only ways would be:
- To delete the page
- To unassign it

Review welcome.

This fixes #150

### DIFF
--- a/meta/Search.php
+++ b/meta/Search.php
@@ -280,6 +280,8 @@ class Search {
         $this->result_pids = array();
         $result = array();
         $cursor = -1;
+        $pageidAndRevOnly = !(count($this->columns) > 2 || $this->columns[0]->getTid() ||
+            !empty($this->columns[1]) && $this->columns[1]->getTid());
         while($row = $res->fetch(\PDO::FETCH_ASSOC)) {
             $cursor++;
             if($cursor < $this->range_begin) continue;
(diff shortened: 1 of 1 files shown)


meta/Search.php
```diff
@@ -529,12 +531,11 @@ public function findColumn($colname) {
...
      * @param Value $value
-     * @param string $pid
      * @return bool
      */
-    protected function isEmptyValue(Value $value, $pid) {
-        if($value->isEmpty()) return true;
-        if(is_a($value->getColumn()->getType(), Page::class) && $value->getRawValue() == $pid) return true;
+    protected function isEmptyValue(Value $value) {
+        if ($value->isEmpty()) return true;
+        if ($value->getColumn()->getTid() == 0) return true;
```
splitbrain: hah. clever!

meta/Search.php
```diff
@@ -280,6 +280,8 @@ public function execute() {
         $this->result_pids = array();
         $result = array();
         $cursor = -1;
+        $pageidAndRevOnly = !(count($this->columns) > 2 || $this->columns[0]->getTid() ||
+            !empty($this->columns[1]) && $this->columns[1]->getTid());
```
splitbrain: What if I select %pageid%, %title%, %lastupdate%? I think the more general way would be to check if all columns have a Tid of 0.

meta/Search.php
```diff
@@ -280,6 +280,9 @@ public function execute() {
         $this->result_pids = array();
         $result = array();
         $cursor = -1;
+        $pageidAndRevOnly = array_reduce($this->columns, function ($pageidAndRevOnly, Column $col) {
+            return $pageidAndRevOnly && ($col->getTid() == 0);
+        }, true);
```
splitbrain: nice use of array_reduce.

PR #166, opened 2016-08-24
Always use raw values for editors

### LINKED ISSUE #140: data in multi page fields lost, when doing preview

splitbrain:
This is another attempt to fix #140. The valueEditor() and multiValueEditor() now always expect raw values to be passed. The Value class was adjusted accordingly. It now allows to set the raw value
(which is what we get from POST). If done, the Value object is treated as a rawonly Value that will throw an exception when you try to access value or displayvalue.

replaces #159


micgro42:
I like it. It makes the code simpler. :+1:

PR #165, opened 2016-08-24
Validation Refactor

splitbrain:
As commented at https://github.com/cosmocode/dokuwiki-plugin-struct/pull/147#issuecomment-240680877 this refactors the Validation mechanism, to avoid reinstantiating the access table classes.

PR #167, opened 2016-08-24
Show new row and reset form in Lookup Editor

splitbrain:
This is based on top of #165. This adds the last missing features to the lookup editor. When a new row is added, that row is now rendered and dynamically added to the end of the table and the from is reset to the initial state.

### DIFF
--- a/script/LookupEditor.js
+++ b/script/LookupEditor.js
@@ -15,6 +15,11 @@ var LookupEditor = function ($table) {
         $table.find('tr').each(function () {
             var $me = jQuery(this);
 
+            // already added here?
+            if ($me.find('th.action, td.action').length) {
+                return;
+            }
+
             // empty header cells
             if ($me.parent().is('thead')) {
                 $me.append('<th class="action"></th>');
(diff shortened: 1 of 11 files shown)


script/LookupEditor.js
```diff
@@ -63,25 +69,35 @@ var LookupEditor = function ($table) {
...
-                    // FIXME
+                .done(function (data) {
+                    $table.find('tbody').append(data);
+                    addDeleteRowButtons(); // add the delete button to the new row
+                    addForm(formdata); // reset the whole form
                 })
                 .fail(function (xhr) {
-                    // FIXME
-                    alert(xhr.responseText)
+                    $errors.text(xhr.responseText).show();
```
micgro42: If I remember correctly, then that response text might be html. So should we set it with `html()`?

splitbrain: See https://github.com/cosmocode/dokuwiki-plugin-struct/blob/lookups/action/lookup.php#L69 I send a text/plain response with the exception text.

Issue #59, opened 2016-03-08
CSV export


splitbrain:
It would be nice if users could export the data of an aggregation as CSV.

stvoigt:
See table2csv-plugin as an example (https://www.dokuwiki.org/plugin:table2csv)

PR #194, opened 2016-11-02
Export Aggregations as CSV #59

splitbrain:
This implements exporting Data from aggregations as CSV for use in other applications like Excel or OpenOffice.

The export Option is enabled by default, but can be disabled per aggregation via ``csv: 0``

### DIFF
--- a/_test/ConfigParser.test.php
+++ b/_test/ConfigParser.test.php
@@ -77,6 +77,7 @@ class ConfigParser_struct_test extends StructTest {
                         1 => false,
                     )
                 ),
+            'csv' => true,
         );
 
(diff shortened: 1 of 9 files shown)


splitbrain:
One thing that's a bit bad currently, is that the renderer to generate the CSV output only handles a small subset of possible syntax. Eg. images won't show up in the export. The same is true for whatever plugins are used in a Wiki type field.

The proper way to solve this would be basing this renderer on a generic plain text (or wiki source) renderer. Unfortunately we don't have such a thing in the wiki core.

We do have a wiki source renderer in the edittable plugin. Would it be worth to extract it from there and move it to the core and then use it here?

PR #205, opened 2016-11-10
Fix datetime entry

### LINKED ISSUE #204: Bug: DateTime-Type does not accept input
When you want to change a DateTime manually (had prefil-today first), it is not possible to make a time input. The input type does not accept a space after the date or a : between hours/minutes/seconds. I am not able to hit the space or : key to get an input. Only numbers and - work.
zwischenablage01

micgro42:
With this the correct class will be set in all the scenarios (page-edit,
inline, lookup). Then changing the date via the datepicker will always
generate time-template as defined in script/EntryEditor.js

However there are still two other points to fix/discuss:
- [ ] this fix currently duplicates code from `types/Date.php`. We could use their `valueEditor()` and then adjust the returned html, but that may not be less complex either.
- [ ] The javascript that generates that "time-template" does not respect the settings made in the schema-editor for that field: [L34](https://github.com/cosmocode/dokuwiki-plugin-struct/blob/1707798d179f1fd7546216573b23820f8390a31f/script/EntryEditor.js#L34)

Fixes #204

### DIFF
--- a/types/DateTime.php
+++ b/types/DateTime.php
@@ -36,10 +36,15 @@ class DateTime extends Date {
      * @return string html
      */
     public function valueEditor($name, $rawvalue) {
+        $name = hsc($name);
+        $rawvalue = hsc($rawvalue);
+
         if($this->config['prefilltoday'] && !$rawvalue) {
             $rawvalue = date('Y-m-d H:i:s');
         }
-        return parent::valueEditor($name, $rawvalue);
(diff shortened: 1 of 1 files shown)


splitbrain:
both things are fine IMHO. we can live with this little duplicated code. Editing the field happens always with a standardized format. The configuration only changes the output format.

PR #211, opened 2016-11-14
fix: ajax-append the 1st row for lookups as well

### LINKED ISSUE #207: Bug: LookupTable will not refresh with first entry
When filling a lookup table the first time (enter the first entry), the table will not refresh itself. You have to reload the whole page to see your entries. After that reload, the page will be refreshed automatically after the next entry.

micgro42:
Empty lookup-aggregations do not have a body on page-load.

This fixes SPR-694 and it fixes #207

### DIFF
--- a/script/LookupEditor.js
+++ b/script/LookupEditor.js
@@ -92,7 +92,8 @@ var LookupEditor = function (idx, table) {
                 $form.serialize()
             )
                 .done(function (data) {
-                    $table.find('tbody').append(data);
+                    const $tbody = $table.find('tbody').length ? $table.find('tbody') : jQuery('<tbody>').appendTo($table);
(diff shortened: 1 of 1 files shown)


script/LookupEditor.js
```diff
@@ -92,7 +92,8 @@ var LookupEditor = function (idx, table) {
                 $form.serialize()
             )
                 .done(function (data) {
-                    $table.find('tbody').append(data);
+                    const $tbody = $table.find('tbody').length ? $table.find('tbody') : jQuery('<tbody>').appendTo($table);
```
splitbrain: This calls $table.find('tbody') twice which should be avoided. I'm also not a fan of the tertiary short hand if-clause.

PR #213, opened 2016-11-15
Fix: Do not render dynamic filters on pdf export

### LINKED ISSUE #169: don't show dynamic filters in print or export
It makes no sense to display them when they can't be used

micgro42:
Do not render the dynamic filters, because mpdf 6, which creates the pdf, is unable to parse `display: none` neither on the `tr` nor on the `th` or `td` tags and hence removing those filters with CSS doesn't work.
This might improve with mpdf 7.

This fixes #169 and fixes SPR-689

### DIFF
--- a/meta/AggregationTable.php
+++ b/meta/AggregationTable.php
@@ -280,7 +280,10 @@ class AggregationTable {
     protected function renderDynamicFilters() {
         if($this->mode != 'xhtml') return;
         if(!$this->data['dynfilters']) return;
-        global $conf;
+        global $conf, $ACT;
+        if ($ACT == 'export_pdf') {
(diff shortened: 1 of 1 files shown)


meta/AggregationTable.php
```diff
@@ -280,7 +280,10 @@ protected function isDynamicallyFiltered() {
     protected function renderDynamicFilters() {
         if($this->mode != 'xhtml') return;
         if(!$this->data['dynfilters']) return;
-        global $conf;
+        global $conf, $ACT;
+        if ($ACT == 'export_pdf') {
```
micgro42: And it's a cleaner solution ✔️

Issue #193, opened 2016-10-17
Update page references when moving pages


psy-q:
Currently moving a page will break all references to that page in any fields of type "page" on any other page. For example, let's assume this schema:

Schema name: infrastructure
Field name: depends-on
Type: Page

Let's assume two pages that use this schema:
- infrastructure:www.example.com
- infrastructure:dns.example.com

On www.example.com, enter dns.example.com in the "depends-on" field. Then rename dns.example.com to ns1.example.com. You will see that the reference to dns.example.com on www.example.com is not renamed, so that page link points to a non-existing page.

splitbrain:
This is surprisingly complex. We need to figure out in which fields are pages stored. However a field may have changed the type during it's lifetime. So we need to change the page values for the fields when they were holding a page but not for when they held something else. Additionally complex: Lookup types may reference a page or an id of a lookup schema.

Got to find a way to make this work in a performant way.

splitbrain:
Alternatively the much simpler solution would be to only adjust current values and keep historic values as is. @micgro42 what do you think?

micgro42:
Well the move-plugin does not move the old attic versions of a page to the new name, so they are still available under the old name. This would mean that only changing the current link would be consistent with that behaviour.

**But:**
* Are we linking to the old revisions of pages in page-fileds when viewing old revisions of a page with struct data?
* The move plugin removes the changelog of the old page, hence the old pages may not be easily navigateable. Though this me be kind of a bug of the move plugin.

splitbrain:
note to myself: I think we'll need a latest field in the multi_* tables to make this easier.

splitbrain:
We should also handle Media moves for the Media type.

PR #214, opened 2016-11-15
Update page and media references on move

splitbrain:
This implements #193 - page and media references are now rewritten when the referenced objects are move using the move plugin. Only the most recent version of the data is adjusted.


action/move.php
```diff
@@ -8,56 +8,186 @@
...
+                    }
+                } else {
+                    switch(get_class($col->getType())) {
+                        case Media::class:
+                            $this->updateColumnID($schema, $col, $old, $new);
+                            break;
+                    }
+                }
+            }
+        }
```
splitbrain: I'm not too happy with this loop. It has a lot of nesting levels, but I'm not sure how to make better. Suggestions?

PR #218, opened 2016-11-17
Add `%lasteditor%` column

micgro42:
With the column `%lasteditor%` one can display who last changed a page.

Todo
------
- [x] ~~Fix~~ Adjust %lastupdated% to show when the page, not its struct data, was last changed
- [x] Needs tests


meta/Page.php
```diff
@@ -0,0 +1,159 @@
...
+
+        $this->updateEntry('lasteditor', $lastEditor);
+
+        $this->lasteditor = $lastEditor;
+    }
+
+    /**
+     * Sets a new title in the database;
+     *
+     * @param string|null $title set null to derive from PID
```
splitbrain: wrong function comment

meta/Page.php
```diff
@@ -0,0 +1,159 @@
...
+        $this->updateEntry('title', $title);
+
+        $this->title = $title;
+    }
+
+    /**
+     * Sets a new title in the database;
+     *
+     * @param string|null $title set null to derive from PID
+     */
```
splitbrain: wrong function comment

meta/Page.php
```diff
@@ -0,0 +1,159 @@
...
+        if($lastrev === null) {
+            $lastrev = 0;
+        }
+
+        $this->updateEntry('lastrev', $lastrev);
+
+        $this->title = $lastrev;
+    }
+
+    protected function updateEntry($column, $entry) {
```
splitbrain: missing doc block

meta/Page.php
```diff
@@ -0,0 +1,159 @@
...
+
+    protected function loadLastRevision() {
+        $sql = "SELECT lastrev FROM titles WHERE pid = ?";
+        $res = $this->sqlite->query($sql, $this->pid);
+        $lastrev = $this->sqlite->res2single($res);
+        $this->sqlite->res_close($res);
+
+        if($lastrev === false) return null;
+        return $lastrev;
+    }
```
splitbrain: I think it would be more effective if this class would always load all three values from the database. This way when I request a title first and then the editor, it would make only one database request.

action/title.php
```diff
@@ -39,12 +39,26 @@ public function handle_meta(Doku_Event $event, $param) {
...
+                $page->setLastRevision(null);
+            }
+
+            if(!blank($event->data['current']['last_change']['user'])) {
+                $page->setLastEditor($event->data['current']['last_change']['user']);
+            } elseif (!blank($event->data['current']['last_change']['ip'])) {
+                $page->setLastEditor($event->data['current']['last_change']['ip']);
+            } else {
+                $page->setLastEditor(null);
             }
```
splitbrain: The above create three database queries to store the data. I think it makes more sense to have something like this:
```
$page->setTitle(...);
$page->setLastEditor(...);
$page->setLastRevision(...);
$page->save();
```

eg. move the save mechanism out of the set methods.

meta/UserColumn.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param bool $enforceSingleColumn ignored
+     * @return string
+     */
+    public function getFullColName($enforceSingleColumn = true) {
+        $col = 'titles.'.$this->getColName($enforceSingleColumn);
+        return $col;
+    }
+
+    /**
+     * @return string always '%lastupdate%'
```
splitbrain: wrong comment

splitbrain:
Looks good. Do you plan on adding any tests or should I merge it?

micgro42:
I'm currently writing some tests, between interruptions. Should be done today, so you can merge tomorrow.

micgro42:
Actually I will only push one of the two tests today, since getting a test for `%lastupdate%` right is tricky. SQLite uses the localtime of the respective system, however PHP may or may not use that time or something else or UTC.

There is a [function](https://secure.php.net/manual/en/function.date-default-timezone-set.php#113864) that reads the timezone from the system and adjusts PHP's setting accordingly, but I'm unsure if or where it would make sense to actually use ist.

Issue #90, opened 2016-04-18
Support other aggregation types


iainhallam:
In the original data plugin, as well as a table, you can pull out a list or cloud; are these likely to be supported in struct?

splitbrain:
Yes, implementing other aggregation types was planned. However since those other types are rarely used in the data plugin they had very low priority for the struct plugin and weren't implemented in the first development cycle.

PRs welcome

regul8or:
I second this feature. Just when I started to use dokuwiki as a replacement for delicious and evernote, I found table aggregation too restrictive for this purpose.

More info for the use case here: https://forum.dokuwiki.org/post/53314

eivdas:
I held off the move to struct because I need datalist, as well as datacloud...

Laynee:
I've built a music database for all my favourite albums. It's awesome with de data plugin because you can generate a tag cloud for all the genres in your collection (Art Rock, Jazz, Memphis Blues...). You instantly see witch genres are the most popular, and it's beautiful ! So I stay with the data plugin for now. 

A tag cloud works really well for all "collections" databases : movies, music, video games, links... There are so much use cases !

splitbrain:
list and cloud have been implemented

PR #227, opened 2016-11-22
Add list aggregation

micgro42:
The goal of this pull-request is to implement the functionality of the data plugin's [`datalist`](https://www.dokuwiki.org/plugin:data#data_list_output) output style.

See #90 and see SPR-710

ToDo
-------
- [x] implement `sepbyheaders`-option?
- [x] ~~tests?~~ (there is not really meaningful business logic here
- [x] remove table-specific functionality and optionally show an error instead

### DIFF
--- a/meta/AggregationList.php
+++ b/meta/AggregationList.php
@@ -138,7 +138,7 @@ class AggregationList {
             }
             $value->render($this->renderer, $this->mode);
             if ($column < $this->resultColumnCount) {
-                $this->renderer->doc .= ' ';
+                $this->renderer->cdata(' ');
             }
             if ($this->mode == 'xhtml') {
                 $this->renderer->doc .= '</div>';
--- a/meta/AggregationList.php
+++ b/meta/AggregationList.php
@@ -72,13 +72,17 @@ class AggregationList {
 
         $this->startScope();
 
-        $this->renderer->doc .= '<ul>';
+        $this->renderer->listu_open();
 
         foreach ($this->result as $result) {
+            $this->renderer->listitem_open(1);
+            $this->renderer->listcontent_open();
             $this->renderListItem($result);
+            $this->renderer->listcontent_close();
+            $this->renderer->listitem_close();
         }
 
-        $this->renderer->doc .= '</ul>';
+        $this->renderer->listu_close();
 
         $this->finishScope();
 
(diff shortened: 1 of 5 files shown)


style.less
```diff
@@ -235,6 +235,21 @@ form.struct_newschema {
  * Aggregation
  */
 .dokuwiki .structaggregation {
+
+    &.listaggregation {
+        > ul li {
+            div {
```
splitbrain: the above three lines could be joined into one, decreasing the nesting

syntax/list.php
```diff
@@ -6,27 +6,36 @@
...
+use dokuwiki\plugin\struct\meta\SearchConfig;
+use dokuwiki\plugin\struct\meta\StructException;
+
 // must be run within Dokuwiki
 if (!defined('DOKU_INC')) die();
 
 class syntax_plugin_struct_list extends DokuWiki_Syntax_Plugin {
+
+    /** @var string which class to use for output */
+    protected $tableclass = AggregationList::class;
```
splitbrain: again, there's only one class. No need to define it here.

splitbrain: OTOH it seems this class and the cloud syntax class duplicate a bit of code from the table and lookup classes. So maybe it's worth to let them inherit and only overwrite a few things?

meta/AggregationList.php
```diff
@@ -0,0 +1,139 @@
...
+            $type = 'struct_' . strtolower($value->getColumn()->getType()->getClass());
+            $this->renderer->doc .= '<div class="' . $type . '">';
+            $value->render($this->renderer, $this->mode);
+            if ($column < $this->resultColumnCount) {
+                $this->renderer->doc .= ' ';
+            }
+            $this->renderer->doc .= '</div>';
+        }
+
+        $this->renderer->doc .= '</div></li>';
```
splitbrain: this whole function seems to assume a XHTML renderer but never makes sure that is the case.

syntax/list.php
```diff
@@ -53,9 +57,34 @@ public function connectTo($mode) {
...
+     * Checks for options that do not work in a list aggregation
+     *
+     * @param array $config
+     */
+    protected function checkForInvalidOptions($config) {
+        $illegal = ['dynfilters', 'summarize', 'rownumbers', 'widths', 'summary'];
+        foreach ($illegal as $illegalOption)
+        if (!empty($config[$illegalOption])) {
+            throw new StructException('illegal option', $illegalOption);
+        }
```
splitbrain: I guess we need this for the other syntax modes. Eg. sepbyheaders is not available for table or lookup.

splitbrain:
great work!

PR #230, opened 2016-11-23
Add cloud aggregation similar to data-plugin's datacloud

micgro42:
These two classes having the same name caused an error that on some
pages: after saving the page, the page itself remained blank.

The error-log contained the following:
`
[Mon Nov 28 09:05:52 2016] [error] [client 91.65.183.141] PHP Fatal error:  Cannot use dokuwiki\\plugin\\struct\\types\\Page as Page because the name is already in use in [...]/lib/plugins/struct/meta/Search.php on line 7, referer: [...]
`

This fixes SPR-720


meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+        $this->result = $this->searchConfig->execute();
+        $this->resultCount = $this->searchConfig->getCount();
+        $this->helper = plugin_load('helper', 'struct_config');
+
+        $this->max = $this->result[0]['count'];
+        $this->min = end($this->result)['count'];
+    }
+
+    /**
+     * Create the table on the renderer
```
splitbrain: wrong comment

meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+        // remove identifier from renderer again
+        if(isset($this->renderer->info['struct_cloud_hash'])) {
+            unset($this->renderer->info['struct_cloud_hash']);
+        }
+        // wrapping div
+        if($this->mode != 'xhtml') return;
+        $this->renderer->doc .= '</div>';
+    }
+
+    protected function renderResult($result) {
```
splitbrain: I thinks this would be better named ``renderTag()``

meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+     * Adds additional info to document and renderer in XHTML mode
+     *
+     * @see finishScope()
+     */
+    protected function startScope() {
+        // unique identifier for this aggregation
+        $this->renderer->info['struct_cloud_hash'] = md5(var_export($this->data, true));
+        // wrapping div
+        if($this->mode != 'xhtml') return;
+        $this->renderer->doc .= "<div class=\"structaggregation cloudaggregation\">";
```
splitbrain: do we need both classes? do we reuse structaggregation styles? cloudaggregation would probably better named structcloud.

meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+         */
+        $value = $result['tag'];
+        $count = $result['count'];
+        $weight = $this->getWeight($count, $this->min, $this->max);
+        $type = strtolower($value->getColumn()->getType()->getClass());
+        if ($value->isEmpty()) {
+            return;
+        }
+
+        $raw = $value->getRawValue();
```
splitbrain: I think this should be getDisplayValue()

meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+        if (is_array($raw)) {
+            $raw = $raw[0];
+        }
+        $schema = $this->data['schemas'][0][0];
+        $col = $value->getColumn()->getLabel();
+        $this->renderer->doc .= '<li><div class="li">';
+        $this->renderer->doc .= "<div data-weight='$weight' data-count='$count' class='cloudtag $type'>";
+        //$value->render($this->renderer, $this->mode);
+        $this->renderer->internallink("?flt[$schema.$col*~]=" . urlencode($raw),$raw);
```
splitbrain: the target page should be configurable (defaulting to the current page)

meta/AggregationCloud.php
```diff
@@ -0,0 +1,157 @@
...
+        if ($column < $this->resultCount) {
```
splitbrain: where is ``$column`` set?

script/effects.js
```diff
@@ -0,0 +1,27 @@
...
+
+        if (bt < at) {
+            return -1;
+        }
+
+        return 0;
+    });
+
+    $taglis.detach().appendTo($taglist);
+})();
```
splitbrain: Why is all this done in JavaScript? I don't see a reason why this cloud couldn't be created with appropriate inline styles and correctly sorted within PHP.

meta/AggregationCloud.php
micgro42: Mh, true. However there is the edge-case that display-value is empty for page-types where the page never existed or raw-value is not a valid page-id (for example because the type was changed from `Text` to `Page`).

Moreover I would actually like to render _some_ of the types, for example colors and images. `#12a6b3` is just not that useful as a tag.

micgro42:
I'm really not sure how checking one of the checkboxes in the initial post caused it to display the text of another, closed pull-request, while still saying it was edited 3 days ago. Especially since the pull request, which's text is now displayed in the first post, has only been created today.

micgro42:
before
after

splitbrain:
As discussed, cropping to image squares would be the better solution for images in clouds.

meta/AggregationCloud.php
```diff
@@ -0,0 +1,240 @@
...
+        if (is_array($tagValue)) {
+            $tagValue = $tagValue[0];
+        }
+        $filter = "flt[$schema.$col*~]=" . urlencode($tagValue);
+        $linktext = $tagValue;
+
+        if ($type == 'Color') {
+            $url = wl($target, $filter);
+            $style = "background-color:$tagValue;display:block;height:100%";
+            return "<a href='$url' style='$style'></a>";
```
splitbrain: this is not renderer agnostic, a fallback for non-xhtml renderers may be needed

splitbrain:
The cloud class currently implements a few special cases for certain types (color, media). I wonder if we should somehow abstract this into the Type classes itself. Plugins may want to do their own handling.

micgro42:
- [ ] implement `getSortString()` in `AbstractBaseType` and overwrite in `Color`
- [ ] implement `renderAsLinkText()` in `AbstractBaseType` and overwrite in `Color` and `Media`

micgro42:
Apparently cropping is only for internal images simple. External images may need javascript or must be rendered as CSS `background-image`. See http://stackoverflow.com/questions/15167545/how-to-crop-a-rectangular-image-into-a-square-with-css

splitbrain:
I would say we handle it as we do in "normal" DokuWiki pages. Images get cropped when they are local or conf[fetchsize] is set high enough to cache the external images. other external images are simply changed to the desired size without respecting the aspect ration (possibly distorting them).

Adding a javascript+background-image fallback should be considered for DokuWiki core though.

types/Media.php
```diff
@@ -111,4 +111,24 @@ public function valueEditor($name, $rawvalue) {
...
+     * @inheritDoc
+     */
+    public function renderTagCloudLink($value, \Doku_Renderer $R, $mode, $page, $filter, $weight) {
+        $media = $this->displayValue($value);
+        if ($mode == 'xhtml' && $this->getConfig()['mime'] == 'image/') {
+            $url = wl($page, $filter);
+            $image = ml($media, ['h' => $weight, 'w' => $weight]);
+            $R->doc .= "<div style=\"height:{$weight}px; width:{$weight}px\">";
+            $R->doc .= "<a href='$url' class='struct_image' style='background-image:url(\"$image\")' title='$media'>";
+            $R->doc .= "<span class='hide_text'>$media</span>";
```
splitbrain: this span isn't really needed, the ``alt`` attribute should be enough. ``$media`` should be esacaped (``hsc``)

meta/AggregationCloud.php
```diff
@@ -0,0 +1,206 @@
...
+        if (!empty($this->data['target'])) {
+            $target = $this->data['target'];
+        } else {
+            global $INFO;
+            $target = $INFO['id'];
+        }
+
+        $tagValue = $value->getDisplayValue();
+        if (empty($tagValue)) {
+            $tagValue = $value->getRawValue();
```
micgro42: There is an edge-case for `page` types with `usetitles` set and pages that never existed or uncleaned data after the column has been converted to `page` type. But maybe that is a bug and should be fixed?

meta/SearchCloud.php
```diff
@@ -0,0 +1,103 @@
...
+        return $QB->getSQL();
+    }
+
+    /**
+     * Execute this search and return the result
+     *
+     * The result is a two dimensional array of Value()s.
+     *
+     * This will always query for the full result (not using offset and limit) and then
+     * return the wanted range, setting the count (@see getCount) to the whole result number
```
micgro42: True, but we also would need to add a `addLimit()` method to `meta\QueryBuilder` which may create some ambiguity over which approach to take.

types/Media.php
micgro42: But then the `<a>`-tag would be empty?

meta/AggregationCloud.php
splitbrain: Yeah I think we should fix that in the Page type.

meta/SearchCloud.php
splitbrain: hmm would we need it in the QueryBuilder? After all the final SQL is build in Search(Cloud) anyway and there's an addLimit() already.

meta/AggregationCloud.php
splitbrain: I opened #247 for that

splitbrain:
great work as well! thank you!

PR #231, opened 2016-11-28
Resolve conflict between type\Page and meta\Page

micgro42:
These two classes having the same name caused an error that on some
pages: after saving the page, the page itself remained blank.

The error-log contained the following:
`
[Mon Nov 28 09:05:52 2016] [error] [client 91.65.183.141] PHP Fatal error:  Cannot use dokuwiki\\plugin\\struct\\types\\Page as Page because the name is already in use in [...]/lib/plugins/struct/meta/Search.php on line 7, referer: [...]
`

This fixes SPR-720

### DIFF
--- a/_test/Type_Dropdown.test.php
+++ b/_test/Type_Dropdown.test.php
@@ -2,7 +2,6 @@
 
 namespace dokuwiki\plugin\struct\test;
 
-use dokuwiki\plugin\struct\meta\Page;
 use dokuwiki\plugin\struct\test\mock\AccessTable;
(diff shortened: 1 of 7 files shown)


splitbrain:
I think PageMeta is generally the better name for this class anyway, so I'm fine with merging it. Otherwise a proper ``use ... as... `` clause should have resolved this as well. Or one could reference types always as ``type\Foo``

splitbrain:
meta/Page.php should  probably be added to the list of deleted.files

Issue #239, opened 2016-12-01
Feature wish: CSV  import


fthommen:
Since the struct plugin can export CSV files, why not also implement a CSV import?  That would come in handy to replace spreadsheet-based data.  W/o import capability, users will continue to keep legacy data in spreadsheets outside the Wiki.

f.

Laynee:
What a great idea, it would be super handy ! I've got lots of Excel files that I'd love to display / manipulate / filter within the wiki. But entering all the data manually is too time consuming so I never tried. Importing all these data with a click on a button would be awesome.
And when I think about all the people who list things in spreadsheets, and all the software / scripts / apps that can export csv files... well, this feature could be insanely popular !

splitbrain:
This has been implemented.

PR #245, opened 2016-12-05
first go at CSV import #239

splitbrain:
This adds CSV importing for Lookup Schemas. The format is fixed (comma
separator, " encapsulation, \ escape). The first row has to contain
headers matching the Schema's field names. Non matched fields are
ignored.

No validation is done on import. All data is stored as is and will then
bw interpreted by the different types.

### DIFF
--- a/admin/schemas.php
+++ b/admin/schemas.php
@@ -81,7 +81,7 @@ class admin_plugin_struct_schemas extends DokuWiki_Admin_Plugin {
             if(isset($_FILES['csvfile']['tmp_name'])) {
                 try {
                     new CSVImporter($table, $_FILES['csvfile']['tmp_name']);
-                    msg('CSV imported', 1);
+                    msg($this->getLang('admin_csvdone'), 1);
                 } catch(StructException $e) {
                     msg(hsc($e->getMessage()), -1);
                 }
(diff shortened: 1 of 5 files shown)


splitbrain:
I added raw exporting functionality as well.

Theoretically we could support importing pageschema data now as well (requiring a pid column and ignoring all data for non-existing pages). I'm not sure how useful that would be.

micgro42:
Mh, having a general export for pageschemas but not an import doesn't make much sense. We should either have both or neither.
However I find having a reliable API for this more important than having an import&export in the admin-UI.

admin/schemas.php
```diff
@@ -153,6 +176,19 @@ protected function html_json(Schema $schema) {
         $form->addButton('import', $this->getLang('btn_import'));
         $form->addHTML('<p>' . $this->getLang('import_warning') . '</p>');
         $form->addFieldsetClose();
+
+        $form->addFieldsetOpen($this->getLang('csvexport'));
```
micgro42: Using a different lang string here might be more consistent with the other labels on that page. Maybe something like `Export data as CSV` ?

meta/CSVExporter.php
```diff
@@ -0,0 +1,111 @@
...
+     * Create the header
+     *
+     * @param Column[] $columns
+     * @return string
+     */
+    protected function header($columns) {
+        $row = '';
+
+        if(!$this->islookup) {
+            $row .= $this->escape('pid');
```
micgro42: What is the motivation behind escaping _every_ field, whether it needs escaping or not?

splitbrain: I believe this is what Excel and LibreOffice do by default, too. CSV is a confusing file format anyway. Keeping all fields enclosed in quotes makes the file a little bit more consistent and easier to parse for 3rd party software.

PR #247, opened 2016-12-07
make sure displayValue of Page always returns something useful

splitbrain:
As discussed in https://github.com/cosmocode/dokuwiki-plugin-struct/pull/230/files#r91265248


splitbrain:
ah never mind. I just saw faa6f42

PR #257, opened 2017-01-03
Group filters from syntax in SQL

micgro42:
The filters coming from syntax should be grouped together in a subgroup to that OR-filters cannot undo the filters on `latest=1` etc.

Since we cannot know in the `AbstractBaseType::filter` function whether there is already a subgroup for us to use or which to use, the solution was to give the correct subgroup to the filter function directly.

Instead of adding another parameter to the already long list of parameters for that function, I have chosen to give the `QueryBuilderWhere` a reference back to the original `QueryBuilder`.

Fixes SPR-744

### DIFF
--- a/meta/QueryBuilderWhere.php
+++ b/meta/QueryBuilderWhere.php
@@ -18,6 +18,7 @@ class QueryBuilderWhere {
     /**
      * Create a new WHERE clause
      *
+     * @param QueryBuilder $QB The QueryBuilder to which this where-clause belongs
      * @param string $type The type of the statement, either 'AND' or 'OR'
      * @param null|string $statement The statement or null if this should hold sub statments
      */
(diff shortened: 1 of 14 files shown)


meta/QueryBuilderWhere.php
```diff
@@ -12,14 +12,17 @@ class QueryBuilderWhere {
     protected $statement;
     /** @var string */
     protected $type = 'AND';
+    /** @var QueryBuilder */
+    protected $QB;
 
     /**
      * Create a new WHERE clause
      *
```
splitbrain: missing doc block on the new $QB var

micgro42: ✔️

micgro42:
@splitbrain Even after double-checking I cannot find a filter-function in the structstatus plugin. Could you point me to it?

splitbrain:
Oh you're right. it's a select(). Never mind then.

PR #271, opened 2017-02-24
struct output occasionally at wrong place // conflict with log and include plugins

micgro42:
Problem description
--------------------------
Sometimes the instructions for a page's struct data is not rendered at the top of a page, but somewhere else or not at all.

The output is missing when the [log Plugin](https://www.dokuwiki.org/plugin:log) is included on the page and at least one entry has been made. If the log-plugin is followed by an included page, then the output is rendered at the head of this included page.

For an reproducible example see: https://demo.sprintdoc.de/projects:sprintdoc:start

Steps taken
---------------
I wrote some tests for the generated instructions:
  * `output_struct_test::test_output()` tests the base behavior.
  * `output_struct_test::test_log_conflict()` tests the bug described here and currently fails.
  * my first approach to fixing this was to adjust `action_plugin_struct_output` so that only the page that was read first would receive the output instructions. However that caused a bug with pages included by the include-plugin, that were suddenly missing their output of struct data. `output_struct_test::test_include_missing_output` catches that bug.

Possible solutions
-------
- find actual solution that works with both log and include plugin
- fix the problem somehow at the log plugin
- mark struct as incompatible with the log plugin

SPR-634

### DIFF
--- a/action/output.php
+++ b/action/output.php
@@ -38,10 +38,7 @@ class action_plugin_struct_output extends DokuWiki_Action_Plugin {
      * @param $param
      */
     public function handle_output(Doku_Event $event, $param) {
-        global $ID, $ACT;
-        if (act_clean($ACT) !== 'show') {
-            return;
-        }
+        global $ID;
         if(!page_exists($ID)) return;
         $ins = -1;
         $pos = 0;
--- a/action/output.php
+++ b/action/output.php
@@ -20,8 +20,6 @@ if(!defined('DOKU_INC')) die();
  */
 class action_plugin_struct_output extends DokuWiki_Action_Plugin {
 
-    protected $lastread = '';
-
     /**
      * Registers a callback function for a given event
      *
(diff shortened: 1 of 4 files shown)


micgro42:
After discussion yesterday I implemented the fix by adding always the struct-output instruction and deciding during the render step.

During my tests I didn't notice any broken functionality.

action/output.php
```diff
@@ -54,11 +38,11 @@ public function handle_read(Doku_Event $event, $param) {
      * @param $param
      */
     public function handle_output(Doku_Event $event, $param) {
-        global $ID;
-        if($this->lastread != $ID) return; // avoid nested calls
-        $this->lastread = '';
+        global $ID, $ACT;
+        if (act_clean($ACT) !== 'show') {
+            return;
```
splitbrain: Could it happen that the parse step is done during another mode than show? Then the instructions would be missing. I don't think we need this check here, since we're testing during rendering now.

micgro42: Mh, then we need to add this check in render, since otherwise we would get the output for `edit` `admin`, etc. as well.

micgro42:
Aside from my comment above, the syntax file alredy checks for `$ID == $INFO['id']`.

action/output.php
splitbrain: Do we? Under what circumstances would a page containing struct data be rendered in a mode where we do not want the struct output?

syntax/output.php
```diff
@@ -79,9 +79,13 @@ public function handle($match, $state, $pos, Doku_Handler $handler) {
      * @return bool If rendering was successful.
      */
     public function render($mode, Doku_Renderer $R, $data) {
+        global $ACT;
         global $ID;
         global $INFO;
         global $REV;
+        if (act_clean($ACT) !== 'show') {
+            return true;
+        }
```
splitbrain: I'm still unsure if this check is necessary and if it actually could be harmful. What if a different mode triggers xhtml rendering and that is cached then. Like some kind of Ajax plugin or something...

micgro42: Summary of our discussion for documentation:

- `PARSER_HANDLER_DONE` is also called when edit session is parsed/rendered
- We have no reliable way to know which file is currently parsed/rendered
- An alternative might be to test that ID is either identical to the first or last read page and add the instruction if either is true
  - however that is also not a clean solution

=> We will use the currently implemented solution. We will adjust it when we encounter a case where it breaks and/or DokuWiki is changed in a way that gives us the information which page we are currently rendering.

Issue #45, opened 2016-03-02
make schema names translatable


splitbrain:
Maybe we need a config array for the schema itself?

PR #274, opened 2017-02-28
Show translatable schema names on pages

micgro42:
This offers a json-input to set translatable schema-labels. However with this PR we loose the Label explaining the editors field.

SPR-841 #45


meta/Schema.php
```diff
@@ -145,6 +150,36 @@ public function __construct($table, $ts = 0, $islookup = false) {
     }
 
     /**
+     * Add the translatable keys to the configuration
+     *
+     * This checks if a configuration for the translation plugin exists and if so
+     * adds all configured languages to the config array.
+     *
+     * Adapted from @see \dokuwiki\plugin\struct\types\AbstractBaseType::initTransConfig
```
splitbrain: could we use a trait to avoid duplicate code?

micgro42: something like `TranslationUtilities.php`:
```php 
<?php

namespace  dokuwiki\plugin\struct\meta;

trait TranslationUtilities {
    /**
     * Add the translatable keys to the configuration
     *
     * This checks if a configuration for the translation plugin exists and if so
     * adds all configured languages to the config array.
     *
     * Adapted from @see \dokuwiki\plugin\struct\types\AbstractBaseType::initTransConfig
     */
    protected function initTransConfig() {
        global $conf;
        $lang = $conf['lang'];
        if(isset($conf['plugin']['translation']['translations'])) {
            $lang .= ' ' . $conf['plugin']['translation']['translations'];
        }
        $langs = explode(' ', $lang);
        $langs = array_map('trim', $langs);
        $langs = array_filter($langs);
        $langs = array_unique($langs);

        if(!isset($this->config['label'])) $this->config['label'] = array();
        // initialize missing keys
        foreach($langs as $lang) {
            if(!isset($this->config['label'][$lang])) $this->config['label'][$lang] = '';
        }
        // strip unknown languages
        foreach(array_keys($this->config['label']) as $key) {
            if(!in_array($key, $langs)) unset($this->config['label'][$key]);
        }
    }

    /**
     * Returns the translated label for this schema
     *
     * Uses the current language as determined by $conf['lang']. Falls back to english
     * and then to the table name
     *
     * @see \dokuwiki\plugin\struct\types\AbstractBaseType::getTranslatedLabel
     *
     * @return string
     */
    public function getTranslatedLabel() {
        global $conf;
        $lang = $conf['lang'];
        if(!blank($this->config['label'][$lang])) {
            return $this->config['label'][$lang];
        }
        if(!blank($this->config['label']['en'])) {
            return $this->config['label']['en'];
        }
        return $this->table;
    }
}
```

Can our autoloader handle that?

splitbrain: An excellent question. I have no idea :-) Quick googling suggests traits are loaded through the same autoloading mechanism as classes, so it should work.

micgro42:
Re: testing

meta/TranslationUtility.php is already 100% covered by the translation-tests and basic functionality for Schema names is covered by the existing tests as well. So I think extra tests would create more complexity than catching things that are not caught already.

splitbrain:
looks great. can you adjust the docs please?

PR #279, opened 2017-03-14
Add compareValue() to AbstractBaseType & Lookup

micgro42:
This fixes a bug, that caused the $STRUCT.table.field$ filter syntax to be broken for lookup fields.  Struct was comparing the row-id of the argument column to the raw value of the referenced column at the
filtered lookup.

There were multiple options to fix this bug:
1. Add a new function to the AbstractBaseType() that return raw value by default (this is the approach implemented by this commit)
2. Add a function to the lookup type only and use introspection to call it if it exists.
3. Do not compare the values in the referenced columns, but compare only the row-ids

The problem with approach 3: It is not possible to compare such lookup-column to any other column except lookups and this is not what we want.
Approaches 1 and 2 have different trade-offs on where additional code/complexity is created. I decided for approach 1 because it is the cleaner overall approach and we might want to use this functionality for future types as well.

### DIFF
--- a/lang/en/lang.php
+++ b/lang/en/lang.php
@@ -79,7 +79,7 @@ $lang['Exception nolookupmix'] = 'You can not aggregate more than one Lookup or 
 $lang['Exception nolookupassign'] = 'You can not assign Lookup schemas to pages';
 $lang['Exception No data saved'] = 'No data saved';
 $lang['Exception no sqlite'] = 'The struct plugin requires the sqlite plugin. Please install and enable it.';
-$lang['Exception comlumn not in table'] = 'There is no column %s in schema %s.';
+$lang['Exception column not in table'] = 'There is no column %s in schema %s.';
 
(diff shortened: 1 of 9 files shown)


micgro42:
The failed test seems to be due to splitbrain/dokuwiki#1837 and unrelated to this PR.

lang/en/lang.php
```diff
@@ -79,6 +79,7 @@
 $lang['Exception nolookupassign'] = 'You can not assign Lookup schemas to pages';
 $lang['Exception No data saved'] = 'No data saved';
 $lang['Exception no sqlite'] = 'The struct plugin requires the sqlite plugin. Please install and enable it.';
+$lang['Exception comlumn not in table'] = 'There is no column %s in schema %s.';
```
splitbrain: typo: ``comlum`` -> ``column``

PR #280, opened 2017-03-15
Do not break migrations of other plugins

micgro42:
It turns out that the new event `PLUGIN_SQLITE_DATABASE_UPGRADE` is visible to all plugins, whether it is their own database or not. Hence we should check that we only handle our own migration.

SPR-850


splitbrain:
looks good. we should also mention this at the event's documentation

micgro42:
Updated documentation for the sqlite plugin: https://www.dokuwiki.org/plugin:sqlite?do=diff&rev2%5B0%5D=1489090397&rev2%5B1%5D=1489582832&difftype=sidebyside

PR #290, opened 2017-04-24
Wrap the hint in the inline editor in a p tag

micgro42:
For it to be read correctly by screen readers and for better styling, wrap the hint in the inline editor in a `<p>`-tag.

SPR-899

### DIFF
--- a/action/inline.php
+++ b/action/inline.php
@@ -91,9 +91,9 @@ class action_plugin_struct_inline extends DokuWiki_Action_Plugin {
         echo '</label>';
         $hint = $this->column->getType()->getTranslatedHint();
         if($hint) {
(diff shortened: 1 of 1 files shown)


splitbrain:
Why would a screen reader not read the content of a div? Why not using ``<p class="hint">`` only?

micgro42:
After consulting for a second time, a `<p class="hint">` might actually be enough. I will amend the pull request accordingly. The idea of using a `<p>` instead of a `<div>` is that it then gets the basic "amount of text" styles that are purely upon a `<p>` tag. These styles are most of the time useful and then need only to be adjusted for the specific use-case instead of repeated.
Also, it appears, it is not relevant for screen readers after all.

PR #304, opened 2017-06-02
Fix invalid forms

### LINKED ISSUE #292: Multiple <input>-tags in one <label>-tag is invalid HTML
With multi-fields of some types (e.g. color) there will be multiple `<input>`-tags within a single `<label>`-tag. This is not valid HTML.

However, I am currently not sure how to solve this nicely.

micgro42:
Multiple inputs within a single label are invalid HTML. This bug can not only occur in multi-fields, but also in plugin-provided single fields which may contain multiple input-fields, which are then combined by JavaScript.

Thus, we provide an id which every type can assign to the edit-field best suited. For multi-types this is usually the new-input-template.

Fixes #292

----
##### semi-related sidenote:
I noted that the functions `\action_plugin_struct_entry::createForm` and `\action_plugin_struct_entry::makeField` are exact copies of `\action_plugin_struct_edit::createForm` and `\action_plugin_struct_edit::makeField`. Also `\action_plugin_struct_entry::createForm` is apparently only used in tests . Method `\action_plugin_struct_entry::makeField` is also used for lookups, but couldn't that use the functionality of `\action_plugin_struct_edit::makeField`? Or am I missing something?

(This should probably be fixed before this is merged)


types/Checkbox.php
```diff
@@ -62,8 +67,11 @@ public function multiValueEditor($name, $rawvalues) {
                 $checked = '';
             }
 
+            $id = !empty($htmlID) ? "id=\"$htmlID\"" : '';
+            $htmlID = null;
```
splitbrain: Why the assignment to `$htmlID` here? It's not passed by reference and not used anymore further down the function?

types/Color.php
```diff
@@ -56,15 +56,16 @@ public function renderMultiValue($values, \Doku_Renderer $R, $mode) {
...
         $class = 'struct_color';
         if(!preg_match('/^#[a-f0-9]{6}$/', $rawvalue)) {
             // any non-color (eg. from a previous type) should default to the default
             $rawvalue = $this->config['default'];
         }
         $name = hsc($name);
         $rawvalue = hsc($rawvalue);
-        $html = "<input name=\"$name\" value=\"$rawvalue\" class=\"$class\" type=\"color\" />";
+        $id = !empty($htmlID) ? "id=\"$htmlID\"" : '';
+        $html = "<input name=\"$name\" value=\"$rawvalue\" class=\"$class\" type=\"color\" $id />";
```
splitbrain: Those lines (in all of the files) are starting to get ugly. I wonder if it would be better to switch to a `buildAttributes()` approach where where we use associative arrays to prepare the HTML attributes. It would also do the empty($htmlID) check for us.

types/Checkbox.php
micgro42: This is a loop. I only want to give the ID to the first checkbox, all the further checkboxes should not receive an ID. Hence I assign `$htmlID` `null` so that the check `!empty($htmlID)` in the line above fails.

types/Color.php
micgro42: I agree. That would make for nicer code and catch/prevent more errors.

types/Checkbox.php
splitbrain: ah of course. nevermind.

PR #308, opened 2017-06-15
Fix styles for bureaucracy forms

### LINKED ISSUE #307: Bureaucracy forms look very strange when using struct_field
Here is an example:

image

Field labels and left/right alignment is a bit all over the place.

This is done with the following code:

```
<form>
Fieldset "Neues Protokoll erstellen"
Action template templates:@@Vorlage@@ protokolle: "_" 
Thanks    "Protokoll wurde erstellt und kann gefüllt werden"

Static "Bitte die teilnehmenden Gruppen, Vorlage, und Datum der Sitzung wählen:"
struct_field "protokolle.itzgroups" @

select "Vorlage" "protokolle|protokoll_abteilungssitzung|protokoll_betrieb1|protokoll_betrieb2|protokoll_planung|protokoll_entwicklung|protokoll_php_woesi|protokoll_c-sharp_jour_fixe|protokoll_lehrlinge|protokolle_webredaktion|protokolle_redaktion|protokoll_clientrunde|protokoll_leitungsteam|protokoll_lim" @
struct_field "protokolle.date" @

Submit
</form>
```

When using only Bureaucracy fields it works OK, but introducing struct_field seems to confuse the form. This is when using 79cb79f5ac50698f3a34a160333cd5fdf6a4fd03.

micgro42:
The restructuring of the entry label↔input html in #304 has caused the
css for struct fields in bureaucracy forms to break. This commit should fix
this and hence fix #307.

### DIFF
--- a/style.less
+++ b/style.less
@@ -88,7 +88,15 @@
 /**
  * Bureaucracy form adjustments
  */
-.dokuwiki form.bureaucracy__plugin label {
+.dokuwiki form.bureaucracy__plugin div.field {
+    display: block;
+    text-align: right;
+
+    label {
+        display: inline-block;
+        width: 47%;
+    }
+
     span {
         line-height: 2em;
     }
@@ -99,7 +107,6 @@
         // fixes for bootstrap3 template
         color: @ini_text;
         font-size: 100%;
-        text-align: left;
     }
 
     span.input {


psy-q:
It's much better this way. But the input titles appear to be vertically oriented on the bottom:

image

So you'd think the groups in the select field ("Betrieb 1", etc.) have no title at all and that "ITZ-Gruppen" is the title for the help text "Hold CTRL or CMD to select multiple values."

Is this something to do with the struct plugin? With non-struct Bureaucracy fields, they titles appear top-aligned.

PR #311, opened 2017-06-16
Allow display of struct cloud in columns 

micgro42:
The current struct cloud can be very confusing, especially if there are many, many tags. Displaying this as a list with multiple columns might then be more helpful.


splitbrain:
could you add a comparison screenshot?

micgro42:
No, not publicly for privacy reasons. But there is no hurry here.

splitbrain:
instead:

```
ul {
  text-align: center;
  li {
    display: inline
  }
}
```

PR #319, opened 2017-07-11
Add API-method to get data of aggregation

micgro42:


micgro42:
New API documentation: https://www.dokuwiki.org/plugin:struct:remote_api

Issue #331, opened 2017-09-18
Table Aggregation pagination - wrong row numbering


solewniczak:
While using table aggregation with `max:` and `rownumbers: 1` every record page counts from "1" while it should continue.

PR #332, opened 2017-09-18
Table aggregation pagination   wrong row numbering #331

solewniczak:
The feature is implemented but I believe that the logic of row counting should be move from AggregationTable into SearchConfig. This will make more sense to me since the whole pagination is handeled in SearchConfig.


micgro42:
Looks good, but please try to use git commit messages that are more indicative of why you are implementing the change in the respective commit. E.g. `fix row numbers resetting with pagination` would be better than `implement the feature`.

Issue #333, opened 2017-09-18
import csv-data for page schemas


solewniczak:
Implement CSV import for page schemas not only lookup tables.

micgro42:
This should now be implemented. ✔️

PR #334, opened 2017-09-19
Import csv data for page schemas #333

solewniczak:
Some important notes:
1. The "pid" column must be present in a CSV file and it's used to bound the data to a proper wiki page.
2. If the page with given "pid" not exists, we omitt the row, send the message to user but the import continues.
3. We use REPLACE instead of INSERT in SQL queries. If the page has some data already bounded, it's replaced with the CSV version. That differs from lookup schemes behavior.
4. The "rev" column is omitted in CSV file (even if present). We always REPLACE data of the latest page revision.

No unit tests were created.


micgro42:
fixes #333

micgro42:
Works for me :+1: :shipit:

PR #339, opened 2017-09-22
Replace lookup placeholders by value not by the row id

### LINKED ISSUE #323: Rowid instead of DisplayValue of lookup-field replaced into Bureaucracy template
### Steps to reproduce:
1. create a page-schema that has a lookup field.
2. create a bureaucracy template that has a placeholder for that field
3. create a page with a bureaucracy form using that template and selecting a value for that field

**expected result:** the placeholder is replaced with the selected value
**actual result:** the placeholder is replaced with the row id corresponding to the selected value

solewniczak:
While creating a new page using bureaucracy template action together with struct provided data, we should replace the lookup-field's placeholders by the value selected by the user. It's much more sensible.

Fixes #323

### DIFF
--- a/_test/Bureaucracy.test.php
+++ b/_test/Bureaucracy.test.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace dokuwiki\plugin\struct\test;
+
+use dokuwiki\plugin\struct\meta;
+use dokuwiki\plugin\struct\meta\AccessTable;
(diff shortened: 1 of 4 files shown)


micgro42:
I think this is the right way to do it 👍

micgro42:
Works for me 🎉

PR #353, opened 2017-11-27
Feature: Configurable default values for fields

andreasferber:
This PR adds a new "default" configuration option to fields that can be used to set a default value. The default is applied whenever no existing schema data is found for a page (for example a new page or a newly assigned schema). This also means that when adding a new field with default to a schema, the default value will not apply to already existing pages using this schema.

In addition, the default is used when adding struct fields to a bureaucracy plugin form.

If you merge this PR, please let me know once you have made a release containing the change available on dokuwiki.org, so that I can add the new option to the documentation at [https://www.dokuwiki.org/plugin:struct:type](url).

### DIFF
--- a/helper/field.php
+++ b/helper/field.php
@@ -82,7 +82,15 @@ class helper_plugin_struct_field extends helper_plugin_bureaucracy_field {
         }
 
         // output the field
-        $value = new Value($this->column, $this->opt['value']);
+        $rawvalue = $this->opt['value'];
+        if (empty($rawvalue)) {
+            $rawvalue = $this->column->getType()->getDefaultValue();
+        }
+        if ($this->column->isMulti()) {
+            $rawvalue = explode(',', $rawvalue);
+            $rawvalue = array_map('trim', $rawvalue);
+        }
+        $value = new Value($this->column, $rawvalue);
         $field = $this->makeField($value, $params['name']);
         $form->addElement($field);
     }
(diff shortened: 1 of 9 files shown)


andreasferber:
The travis check failed because there is a general issue running tests with PHP 7.0 and stable Dokuwiki  due to changes in phpunit. Maybe you should disable this test for now until Dokuwiki has released a new stable version?

meta/AccessTable.php
```diff
@@ -207,14 +207,20 @@ protected function consolidateData($DBdata, $asarray = false) {
...
 
-            // multi val data is concatenated
-            if($col->isMulti()) {
-                $val = explode($sep, $val);
-                $val = array_filter($val);
+                // multi val data is concatenated
+                if($col->isMulti()) {
+                    $val = explode(',', $val);
+                    $val = array_map('trim', $val);
+                }
```
splitbrain, line 223: It seems the else branch logic changed here. The isMulti() check used to be outside the else, now it is inside?

helper/field.php
```diff
@@ -82,7 +82,15 @@ public function renderfield($params, Doku_Form $form, $formid) {
         }
 
         // output the field
-        $value = new Value($this->column, $this->opt['value']);
+        $rawvalue = $this->opt['value'];
+        if (empty($rawvalue)) {
```
splitbrain, line 86: This would overwrite a value of `0` with the default value.

splitbrain (changes requested):
Generally not a bad idea. Needs merge conflict resolving and the logic needs to be double checked.

PR #354, opened 2017-12-07
fix: struct_field in bureaucracy form now supports a default value

solewniczak:
Before that we cannot use bureaucracy default values for struct_field ("=default value" syntax). Now it's possible.


solewniczak:
Related with: #353

moz0817:
It seems so that default values don't work for field types **Lookup** and **Page** (usetitles on). 
Hopefully ist only a little problem, because default values are very useful.

PR #355, opened 2017-12-07
we need to use p_set_metadata rather than p_get_meatadata here

### LINKED ISSUE #337: Struct header not showing up on pages created by Bureaucracy template
A page created using Bureaucracy template with some struct data doesn't display default struct header (created by action_plugin_struct_output). The header becomes visible after next page revision.

solewniczak:
fixes #337

### DIFF
--- a/action/bureaucracy.php
+++ b/action/bureaucracy.php
@@ -155,10 +155,12 @@ class action_plugin_struct_bureaucracy extends DokuWiki_Action_Plugin {
                 );
 
                 // trigger meta data rendering to set page title
-                // expire the cache in order to correctly render the struct header on the first page visit
-                p_get_metadata($id, array('cache' => 'expire'));
+                p_get_metadata($id);
             }
         }
+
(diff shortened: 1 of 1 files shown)


action/bureaucracy.php
```diff
@@ -155,11 +155,13 @@ public function handle_save(Doku_Event $event, $param) {
                 );
 
                 // trigger meta data rendering to set page title
-                // expire the cache in order to correctly render the struct header on the first page visit
-                p_get_metadata($id, array('cache' => 'expire'));
+                p_get_metadata($id);
```
splitbrain: This line seems to not do anything anymore now. Can we remove it?

solewniczak: Yes. I left it because it was present in the original file (before my first commit on this issue). But the current commit doesn't solve the problem completely. Usually it works but for some cases not (I mean that the page don't have the struct headers visible on the first revision). Unfortunately I didn't find the pattern yet.

PR #356, opened 2017-12-07
feature: hidden fields in bueraucracy form from struct fields

solewniczak:
This commit adds extra syntax which allow us to add struct fields to bureaucracy forms as hidden fields:

    struct_fieldhidden "products.product" "=value"

This only make sense with #354


micgro42:
:+1:

Issue #349, opened 2017-11-03
noob-contribution: new type LongText (mix of Text and Wiki, <textarea> for input but without wiki-parser overhead)


kruff-a-lon:
Hi, sorry for being a git noob, this is such a small thing that I don't really want to put in the effort to understand how git (and github) works...

Anyway, I made a new struct-type: LongText, it is a combination of Text and Wiki, in that it does provide a textarea for input but does not put the contents through the dokuwiki-renderer. Hopefully it can be useful for someone that knows how to work with git so that it can be put somewhere more useful than this comment.

----
```
<?php
namespace dokuwiki\plugin\struct\types;

use dokuwiki\plugin\struct\meta\QueryBuilder;
use dokuwiki\plugin\struct\meta\QueryBuilderWhere;

class LongText extends AbstractMultiBaseType {
    use TraitFilterPrefix;

    protected $config = array(
        'prefix' => '',
        'postfix' => '',
    );


    /**
     * Output the stored data
     *
     * @param string|int $value the value stored in the database
     * @param \Doku_Renderer $R the renderer currently used to render the data
     * @param string $mode The mode the output is rendered in (eg. XHTML)
     * @return bool true if $mode could be satisfied
     */
    public function renderValue($value, \Doku_Renderer $R, $mode) {
        $R->cdata($this->config['prefix'] . $value . $this->config['postfix']);
        return true;
    }

    /**
     * Clean line endings
     *
     * @param int|string $rawvalue
     * @return int|string
     */
    public function validate($rawvalue) {
        $rawvalue = rtrim($rawvalue);
        $rawvalue = cleanText($rawvalue);
        return $rawvalue;
    }
    
    /**
     * Use a text area for input
     *
     * @param string $name
     * @param string $rawvalue
     * @param string $htmlID
     *
     * @return string
     */
    public function valueEditor($name, $rawvalue, $htmlID) {
        $rawvalue = formText($rawvalue);
        $params = array(
            'name' => $name,
            'class' => 'struct_'.strtolower($this->getClass()),
            'id' => $htmlID
        );
        $attributes = buildAttributes($params, true);

        return "<textarea $attributes>$rawvalue</textarea>";
    }
}

```

splitbrain:
I think this makes sense. Can you make this a proper pull request please?

PR #360, opened 2017-12-26
Ticket #349 longtext

rottaran:
Created pull request for kruff-a-lon's textarea field. Made the rows and cols of textarea in the Wiki and LongText types configurable.


splitbrain:
Looks good to me. One minor thing - would it make sense to have the Wiki type inherit from the LongText type? It seems there is some duplicated code between the two now.

rottaran:
Excellent idea. Worked well in my manual test.

PR #362, opened 2018-01-03
add div and span fields around inline form to help the autocompletion

### LINKED ISSUE #326: In-line editing of aggregated table causes "ajax.php:67" error.
image
I have an aggregated table above. The title column is a page, and the other linked ones are tags. Auto-complete works fine when editing the form on the respective pages. However, when I try to in-line edit, when the auto-complete triggers I get an error like so...
image
Here is my console log from Chrome Version 60.0.3112.78 (Official Build) (32-bit)...
[rtfm.kojukination.com-1501268309225.txt](https://github.com/cosmocode/dokuwiki-plugin-struct/files/1184018/rtfm.kojukination.com-1501268309225.txt)
I tested this with both CDN jquery loading, and local, same thing happens, but I left it on local for now just in case.

Let me know if I can provide anything else to help troubleshoot.

Thanks,
Steve

rottaran:
fixes #326 

Was caused by following line in EntryEditor.js: `var name = jQuery(this.element[0]).closest('div.field').find('label').first().data('column');`. The `div.field` was missing.

### DIFF
--- a/action/inline.php
+++ b/action/inline.php
@@ -87,15 +87,19 @@ class action_plugin_struct_inline extends DokuWiki_Action_Plugin {
         // output the editor
         $value = $this->schemadata->getDataColumn($this->column);
         $id = uniqid('struct__', false);
+        echo '<div class="field">';
         echo '<label data-column="' . hsc($this->column->getFullQualifiedLabel()) . '" for="' . $id . '">';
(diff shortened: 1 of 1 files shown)


splitbrain:
thanks!

Issue #296, opened 2017-05-10
Better Date-Picker


micgro42:
The current date-picker does not allow directly selecting the year. Currently, selecting the birthday of a 50-year-old person requires 600 clicks on the Back-Button.

cjohnsonuk:
and it doesn't allow selection of time for date time fields...
A "now" button would be nice to include date and time.

stvoigt:
If you work on the date-picker, please check if it works with all other types in one schema. Currently I have one schema where the date picker works and a different one, where it does not work. If you are interested in my schema definition, please let me know. (The picker does not react on selecting the specific date, i.e. nothing happens.)

micgro42:
@stvoigt Yes, your actual schema definitions would be useful. However, what you describe here sounds like a new/different problem.
Your browser is relevant as well.

stvoigt:
okay, i will open a new issue for that...

PR #363, opened 2018-01-04
use HTML5 date and datetime-local inputs with jquery as fallback

rottaran:
Fixes ticket #296. Also made the year easier to edit in the jQuery datepicker.

### DIFF
--- a/script/EntryEditor.js
+++ b/script/EntryEditor.js
@@ -14,39 +14,32 @@ var EntryEditor = function($form) {
     $form.find('.struct .hashint').tooltip();
 
     /**
-     * Attach datepicker to date types.
-     * Only if browser does not support HTML5 date input.
+     * Attach datepicker to date types, if lacking HTML5 support.
      */
-    var testelem = document.createElement('input');
-    testelem.setAttribute("type", "date");
-    if (testelem.type === "text") {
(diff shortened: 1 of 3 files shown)


script/EntryEditor.js
```diff
@@ -14,28 +14,39 @@ var EntryEditor = function($form) {
...
     /**
-     * Attach datepicker to date types
+     * Attach datepicker to date types.
+     * Only if browser does not support HTML5 date input.
      */
-    $form.find('input.struct_date').datepicker({
-        dateFormat: 'yy-mm-dd'
-    });
+    var testelem = document.createElement('input');
+    testelem.setAttribute("type", "date");
```
splitbrain: Hmm, you're setting the type in the HTML of the actual input field - I wonder what `attr()` does return on browsers not supporting it. I would assume that they return `text` as well. If that's the case the check could be done on the real element without the need to create a new element.

rottaran: jquery's `attr()` returns what is written in the HTML code, that is "date", whereas the DOM element `.type` gives "text". Thus, I did not know how to filter this with jquery's `find()`. But '.filter()' should work and I shall simplify the code accordingly.

splitbrain: very elegant solution!

PR #368, opened 2018-01-08
"struct_lookup" action for bureaucracy

solewniczak:
Action allows create struct lookup entries using bureaucracy form. Syntax:
  <form>
  action struct_lookup
  ...
  </form>

<form> can contain any number of struct_field or struct_schema. Data will be saved to respective lookup schemas.

### DIFF
--- a/helper/lookup.php
+++ b/helper/lookup.php
@@ -38,7 +38,8 @@ class helper_plugin_struct_lookup extends helper_plugin_bureaucracy_action {
             if (!$access instanceof AccessTableLookup) continue;
 
             if(!$access->getSchema()->isEditable()) {
-                throw new StructException('lookup save error: no permission for schema');
+                msg('lookup save error: no permission for schema', -1);
+                return false;
             }
             $validator = $access->getValidator($data);
             if($validator->validate()) {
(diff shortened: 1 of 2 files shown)


solewniczak:
Works well with  #356

helper/lookup.php
```diff
@@ -0,0 +1,58 @@
...
+            if(!isset($tosave[$tbl])) $tosave[$tbl] = array();
+            $tosave[$tbl][$lbl] = $field->getParam('value');
+        }
+
+        foreach($tosave as $table => $data) {
+            $access = AccessTable::byTableName($table, 0, 0);
+            if (!$access instanceof AccessTableLookup) continue;
+
+            if(!$access->getSchema()->isEditable()) {
+                throw new StructException('lookup save error: no permission for schema');
```
splitbrain: Does throwing a Struct Exception in the bureaucracy context work? I don't think bureaucracy is catching exceptions. Instead a msg() should be shown and false be returned.

splitbrain:
can you update the docs again? thanks!

solewniczak:
Ok. No problem.

PR #369, opened 2018-01-10
fix: use multi input in bureaucracy forms

solewniczak:
this commit fixes a bug that causes an error when submitting bureaucracy form with struct multi input.

additionally it fixes one minor bug that causes struct lookup fields loosing their selected values when a bureaucracy form was send with an error.

works only with: https://github.com/splitbrain/dokuwiki-plugin-bureaucracy/pull/226

### DIFF
--- a/action/bureaucracy.php
+++ b/action/bureaucracy.php
@@ -87,7 +87,9 @@ class action_plugin_struct_bureaucracy extends DokuWiki_Action_Plugin {
             if(!is_a($field, 'helper_plugin_struct_field')) continue;
             if($field->column->getType()->getClass() != 'Lookup') continue;
 
-            $pid = $field->getParam('value');
+            $value = $field->getParam('value');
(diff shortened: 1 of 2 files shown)


micgro42:
I tried it together with the bureaucracy branch and it worked for me :+1:  . However, I'm hesitant to merge it. Mainly because this would require merging the bureaucracy pr, which is something I do not want to do without tests.
Bureaucracy is very complex and largely untested, hence I'm somewhat afraid of merging changes I don't fully understand even if they are looking good.

I'd love to have this functionality as soon as bureaucracy has better test coverage.

solewniczak:
splitbrain/dokuwiki-plugin-bureaucracy#226 is merged, so this may be merged too.

micgro42:
@solewniczak Thanks for the heads-up. Merged. :+1:

PR #370, opened 2018-01-11
special columns for lookup select fields

solewniczak:
I'm not quite sure if it's more feature or bugfix but i think that special columns of struct schemas should also be available for struct lookup field.


micgro42:
Seems fine :+1:

Issue #145, opened 2016-08-07
Struct data as variables


eilko:
I have added several fields to a page with the use of the Struct plugin. Adding data is working great and so are the aggregated tables.

But I can't get the data of the page itself shown in the way I would like to. I have disabled the show on page feature as I don't want to see all fields and I want to combine them in a new table with some other fields. Secondly I would like to decide myself where to place the table instead of on top like the Struct plugin does now. 

**Feature request:**
Use the data of a page in the fields of the struct plugin as variables in the page. Or a way to decide where to place the Struct table on the page.

micgro42:
You could simply make an aggregation table and filter `%pageid%` with `$ID$`. Then you have full control on what you want to show.

eilko:
@micgro42 your suggestion makes it possible to show the table below the image instead of directly under the page name, and I can choose what to show in the table. But it is a one line table instead of a "normal" table (see example below). Besides that I would like to be able to add a row to the table that is not in the data fields as it will contain a lot of wiki (see field remarks).

<img width="732" alt="screen shot 2016-08-07 at 10 53 19" src="https://cloud.githubusercontent.com/assets/1078401/17461456/35d9a4b6-5c8d-11e6-8423-6acf68f899c3.png">

It would be great if I could make my own table like below, where the @@VARIABLES@@ are struct plugin data of that page. 

|Year:|[[:timeline:@@YEAR@@]]|
|Publication|@@PUBLICATION@@|
|Edition:|@@EDITION@@|
|Remarks:|**This** is some [[:dictionary:|random]] text //to show// the purpose|

iainhallam:
It seems to me that this could be really powerful if you could simply specify a query to come up with a replacement for inline use rather than as a list or table:

    {{struct schemas: publication; cols: year; filter: %pageid% = $ID$;}}

To make this much easier to type, when the schema is set to, e.g., `_self`, the list of included schemas could default to whatever is assigned to the current page, and the filter could add the current page ID as a criterion automatically. In fact, if no schema is specified, this could be the default anyway, so a user would only have to do something like the following to get a single variable:

    {{struct select:year;}}

You could make a syntactic sugar version of that, too:

    {{structvar year}}

You'd still have access to any other function available to the main struct queries, though, so you could select any data value across the whole database to be used in your page.

This could also allow some calculations to be performed like:

    {{struct from: publication; select:AVERAGE(year);}}

But that would probably be a different issue to report once the basic feature was in place!

[...]

PR #386, opened 2018-02-26
Value aggregation type

iainhallam:
This addition addresses #145 and adds an inline syntax following the existing Struct model, plus a shorthand for a single variable; examples follow.

Value from the current page:
```
{{$myschema.myfield}}
```

Value from another page:
```
{{$"ns:second.page".myschema.myfield}}
```

Filters can be added manually following a question mark; to keep flexibility for future features like `summarize`, this means you need to restrict to the page you want yourself. Filters must be in double quotes:
```
{{$$ID$.myschema.myfield ? filter: "myfield < 5" or: "myfield > 5"}}
```

Filtering supports the usual Struct keywords, `filter`, `where`, `filterand`, `and`, `filteror`, `or`, followed immediately by a colon, or the shortcuts `&` and `|`. The value above can be written:
```
{{$$ID$.myschema.myfield ? & "myfield < 5" | "myfield > 5"}}
```


iainhallam:
I think that this check failed due to external factors - is there a way to rerun the tests?

iainhallam:
@splitbrain I believe the build failure was at the time that the stable build didn't work with PHP 7 - is it possible to rerun the tests any way, please?

eilko:
@splitbrain can this request be merged please?

moz0817:
Very useful extension - thank you very much. I have two annotations:

1: It don't works with relative sites like sidebar or pageheader. This might be useful two.
2: If the shown values would be editable - my luck would be unbelievable ;-). Then it's possible to create your own edit forms.

annda:
This looks very good, I would also like to see the inline syntax merged. The new feature probably does not require any test, maybe for the InlineConfigParser

@splitbrain what do you think?

meta/AggregationValue.php
```diff
@@ -0,0 +1,153 @@
...
+        // Load helper
+        $this->helper = plugin_load('helper', 'struct_config');
+    }
+
+    /**
+     * Create the output on the renderer
+     * 
+     * @param  int  $show_not_found  Whether to display the default text for no records
+     */
+    public function render($show_not_found = 0) {
```
splitbrain: It would probably be better to pass a boolean here.

meta/AggregationValue.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->startScope();
+
+        // Check that we actually got a result
+        if ($this->resultCount) {
+            $this->renderValue($this->result);
+        } else {
+            if ($show_not_found) {
+                $this->renderer->cdata($this->helper->getLang('none'));
+            }
```
splitbrain: I wonder if this makes sense as a global setting. Wouldn't it be better if I could specify a default value per use? eg. ``{{$myschema.myfield|default value}}``

meta/InlineConfigParser.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * parses the given inline configuration
+     *
+     * @param  string  $inline
+     */
+    public function __construct($inline) {
+        // To support a fuller syntax, such as user-specified filters, tokenise here
+
+        // Split into components
+        $components = str_getcsv($inline, '.');
```
splitbrain: Cool idea to use the CSV parsing here to support quoting and stuff, but I wonder if that's necessary. Wouldn't a simple `explode` be sufficient?

meta/InlineConfigParser.php
```diff
@@ -0,0 +1,55 @@
...
+
+        // Protect against using single quotes
+        foreach ($components as $component) {
+            if ( substr($component, 0, 1) == "'" ) {
+                // Used single quotes rather than double - will need to rerun CSV extraction
+                $enclosure = "'";
+            }
+        }
+        if ($enclosure == "'") $components = str_getcsv($inline, '.', $enclosure);
```
splitbrain: And here it gets a bit ugly. As far as I can see the enclosure/quoting is only necessary to support page names containing dots. Using a separator that is not allowed in pages could circumvent this problem. ``{{$page.name>schema>value}}`` could work for example.

meta/InlineConfigParser.php
```diff
@@ -0,0 +1,55 @@
...
+            // At least schema and field supplied
+            $lines[] = 'schema: ' . trim($components[0]);
+            $lines[] = 'field: ' . trim($components[1]);
+            $lines[] = 'filter: %pageid% = ' . $GLOBALS['ID'];
+        } 
+
+        // Call original ConfigParser's constructor
+        parent::__construct($lines);
+    }
+}
```
splitbrain: Having some test for this class would be good.

syntax/value.php
```diff
@@ -0,0 +1,136 @@
...
+     * @param string $mode Parser mode
+     */
+    public function connectTo($mode) {
+        /*
+         * Full syntax:
+         * ---- struct value ----
+         * <config>
+         * ----
+         */
+        $this->Lexer->addSpecialPattern('----+ *struct value *-+\n.*?\n----+', $mode, 'plugin_struct_value');
```
splitbrain: I wonder how useful this syntax is. Would you really want to configure a complex filter for a single value? And if so, would you want to do it in this very blocky syntax?

meta/InlineConfigParser.php
```diff
@@ -0,0 +1,55 @@
...
+        if (count($components) == 3) {
+            // At least page, schema and field supplied
+            $lines[] = 'schema: ' . trim($components[1]);
+            $lines[] = 'field: ' . trim($components[2]);
+            $lines[] = 'filter: %pageid% = ' . trim($components[0]);
+        } elseif (count($components) == 2) {
+            // At least schema and field supplied
+            $lines[] = 'schema: ' . trim($components[0]);
+            $lines[] = 'field: ' . trim($components[1]);
+            $lines[] = 'filter: %pageid% = ' . $GLOBALS['ID'];
```
splitbrain: Please don't use the `$GLOBALS` array but import the global properly with the `global` keyword. I also wonder if it wouldn't be better to use `$INFO['id']` so this syntax could be used within a sidebar or include and results would change depending on the current page.

syntax/value.php
```diff
@@ -0,0 +1,136 @@
...
+
+        // Get configuration
+        $show_not_found = $this->getConf('show_not_found');
+
+        try {
+            $search = new SearchConfig($data);
+
+            /** @var AggregationValue $value */
+            $value = new AggregationValue($INFO['id'], $mode, $renderer, $search);
+            $value->render($show_not_found);
```
splitbrain: the above does not need to happen during metadata rendering.

fschrempf:
@iainhallam Do you intend to respond to @splitbrain's review comments above?
It would be great to have this merged and it looks like you were close to bring this over the finish line back in 2018.

meta/AggregationValue.php
iainhallam: This is taken directly from the config array, so will always be given a 0 or 1 depending on the config file. Should I convert a config 0 to Boolean false and config 1 to Boolean true before making use of the value, @splitbrain?

meta/AggregationValue.php
iainhallam: That would be a nice enhancement to raise as an issue once this is merged, I think.

meta/InlineConfigParser.php
iainhallam: That was my first thought, but dots are allowed in page names, so some more intelligent handling needs to be used. Rather than rewriting the logic, I decided to use the built-in.

iainhallam: Oh, reading on to the other comments, this can be dealt with another way...

meta/InlineConfigParser.php
iainhallam: Good shout; the `>` character gets a bit lost on its own in some fonts, so I'm looking at `->` with `explode()`.

syntax/value.php
iainhallam: Is this not where the value would be put into the text of the abstract when rendering metadata? This is quite a complex part of the DokuWiki code and I'm not at all sure I understand it!

meta/InlineConfigParser.php
iainhallam: All other locations in Struct use the dotted notation, and the ugly bit can be avoided by restricting to one type of quote.

syntax/value.php
iainhallam: Removed.

fschrempf:
@iainhallam Thanks a lot for picking up this work. I just tested your latest changes and while I can get the value from the schema to be rendered correctly in normal text, it doesn't work in headings. Is this a known/expected limitation?

```
====== Header1 {{$test.test}} ======

Struct data: {{$test.test}}
```

becomes:
grafik

fschrempf:
One other thing I noted with a rather special use case:

I want to overwrite the meta page title with data from a struct field. So I tested this PR in combination with the [page title plugin](https://www.dokuwiki.org/plugin:pagetitle). The page content rendering is correct, but the meta title shows "Nothing found" when `show_not_found` is enabled. So somehow the struct data is not passed to the meta renderer correctly!?

```
====== Header1 {{"$test.test"}} ======

Struct data: {{$test.test}}

<title>
{{$test.test}}
</title>
```

becomes:

grafik

iainhallam:
> @iainhallam Thanks a lot for picking up this work. I just tested your latest changes and while I can get the value from the schema to be rendered correctly in normal text, it doesn't work in headings. Is this a known/expected limitation?

This is a limitation of DokuWiki - you can't have any syntax inside headers:

https://www.dokuwiki.org/faq:headerlinks

fschrempf:
Ok, for the first example. That's a DokuWiki limitation. What about the second example using the pagetitle plugin? I think that's a different case, as the syntax is actually parsed, only the data is missing in the meta rendering.

meta/InlineConfigParser.php
iainhallam: @splitbrain I've written a test class, but what command line parameters do you use with phpunit to run the tests for just this plugin?

iainhallam:
> Ok, for the first example. That's a DokuWiki limitation. What about the second example using the pagetitle plugin? I think that's a different case, as the syntax is actually parsed, only the data is missing in the meta rendering.

I'll need to take some time to delve into this - I don't feel I have a handle on the metadata parts of DW at the moment.

fschrempf:
> I'll need to take some time to delve into this - I don't feel I have a handle on the metadata parts of DW at the moment.

Sure. I had a quick look for myself, but unfortunately I couldn't get the hang of it either so far. And I'm not even sure if this should be in the scope of this PR or rather something to look at after getting this merged.

Juergen-aus-Zuendorf:
Could you perhaps test if your changes would support to use this variable inside an aggregation as filter, described in issue https://github.com/cosmocode/dokuwiki-plugin-struct/issues/145, comment "[on 2 Apr 2019](https://github.com/cosmocode/dokuwiki-plugin-struct/issues/145#issuecomment-478963174)" ? 

Example:
```
---- struct table ----
schema : fbg_ident, fbg_meta
cols   : fbg_ident.fbg-fkt, fbg_meta.fbg-mtyps
filter : fbg-schlagwort = {{$my_page.fbg_meta.fbg-mtyps}}
----
```

This would be extremely helpful. Thanks a lot in advance

Regards, Juergen

iainhallam:
@Juergen-aus-Zuendorf
> ```
> ---- struct table ----
> schema : fbg_ident, fbg_meta
> cols   : fbg_ident.fbg-fkt, fbg_meta.fbg-mtyps
> filter : fbg-schlagwort = {{$my_page.fbg_meta.fbg-mtyps}}
> ----
> ```

This is already available for the current page using the format `filter : fbg-schlagwort = $STRUCT.fbg_meta.fbg-mtyps$`. See the docs at:

https://www.dokuwiki.org/plugin:struct:filters

My changes essentially add an inline syntax to the Struct configuration parser and an aggregation to display a single value, so this would be a change to the Struct configuration parser logic itself, that would need to add the ability to specify a page in that type of filter. Can you raise a separate issue for that, please?

meta/AggregationValue.php
splitbrain: yeah, I would convert it.

meta/InlineConfigParser.php
splitbrain: see https://www.dokuwiki.org/devel:unittesting#plugins

syntax/value.php
splitbrain: Hmm, I think you might be right. I forgot about that. Maybe give it a quick try by putting the value right at the top of the page and see what ends up in the abstract

splitbrain:
Sorry for the delay in answering your comments. @annda if you're working on this, maybe fix the few little things from above right on merging.

annda:
I find this feature very useful and would like to finally merge it.

I have two questions though:

**serial data:** At this point there is no check on `rid` so the search might in some rare cases return serial data instead of page data. Is this something we should be concerned about? (Probably simply adding `%rowid% = 0` to the filters in `InlineConfigParser` would take care of that.)

**syntax:** The `{{$...}}` pattern is concise and easy to write, but there is no obvious association with the plugin it belongs to. As someone who often needs to troubleshoot, I prefer the earlier version of `{{structvalue ...}}`. Any opinions on that, everyone?

Juergen-aus-Zuendorf:
Although due to the long time span in the meantime, I already worked with iainhallam's extension (thanks for this) and now I would have to adjust the syntax everywhere: 
On the one hand, {{structvalue ...}} is much better for association with the plugin. On the other hand, if you might want to search and replace syntax elements of the function, you can easily search with dokuwiki's search mechanism.

iainhallam:
> One more question: What's about the filtering syntax "`{{$$...`"?

@Juergen-aus-Zuendorf Where's that syntax used; have you got an example?

Juergen-aus-Zuendorf:
Perhaps this is a misunderstanding from myside:

At the beginning of this thread you described the following:
`{{$$ID$.myschema.myfield ? & "myfield < 5" | "myfield > 5"}}`
Therefore, my previous question: "What's about the filtering syntax "{{$$..."?" 
Is it possible to use placeholders within the syntax? And how should it be written in this case?

iainhallam:
> At the beginning of this thread you described the following:
> `{{$$ID$.myschema.myfield ? & "myfield < 5" | "myfield > 5"}}`
> Therefore, my previous question: "What's about the filtering syntax "{{$$..."?"
> Is it possible to use placeholders within the syntax? And how should it be written in this case?

Ah. The configuration is mainly syntactic sugar, so it passes the page ID straight on to the main Struct parser. That syntax is actually the same `{{$` with the ID placeholder from the [main Struct documentation](https://www.dokuwiki.org/plugin:struct:filters#value_placeholders), `$ID$` (meaning _this current page_), as the page name.

It would perhaps be clearer to require a user to specify that they want the query to apply to all pages with something like:

```
/* Current page */
{{$myschema.myfield ? filter: "myfield < 5" or: "myfield > 5"}}

/* Specific page */
{{$other-page.myschema.myfield ? filter: "myfield < 5" or: "myfield > 5"}}

/* Search all pages */
{{$*.myschema.myfield ? filter: "myfield < 5" or: "myfield > 5"}}
```

Juergen-aus-Zuendorf:
I think your suggestion is very good!

Alternatively, I would find it just as good to use the automatism with the placeholders, if this is documented in detail in the plugin description.

And once again my attempt (as syntactic sugar):
Couldn't the filtering likewise support the comparison with a Struct value of another page, e.g. in this form:
`{{$myschema.myfield ? filter: "myfield = {{$other-page.otherschema.otherfield}}"}}`

Best regards
Juergen

iainhallam:
> I think your suggestion is very good!

Will have a go at redoing that part when I can.

> Couldn't the filtering likewise support the comparison with a Struct value of another page, e.g. in this form:
> `{{$myschema.myfield ? filter: "myfield = {{$other-page.otherschema.otherfield}}"}}`

Struct already has a syntax for that, and your bug report #552 should take care of specific pages. I won't be adding it to this aggregation.

annda:
@iainhallam Can you update the documentation?

iainhallam:
Documentation updated at https://www.dokuwiki.org/plugin:struct:aggregation.

PR #390, opened 2018-03-30
Feature %lastsummary% column in page schema

solewniczak:
This commit adds '%lastsummary%' special column for page schemas. The column contains the summary of last page revision.


solewniczak:
What about this commit? Don't you think it may be useful in some cases?

PR #404, opened 2018-07-07
📱 Improve style of struct bureaucracy fields

micgro42:
Especially when it comes to responsive design

This may still need some minor adjustments:

- DokuWiki Template:
  - [ ] on mobile the labels of the struct fields are on the right side, the other labels are on the left
- SprintDoc Template:
  - [ ] the normal labels are bold, the labels of the struct fields are not

### DIFF
--- a/style.less
+++ b/style.less
@@ -93,12 +93,13 @@
     text-align: right;
 
     label {
-        display: inline-block;
-        width: 47%;
+        clear: both; // this is somehow missing in the bootstrap template
     }
 
     span {
         line-height: 2em;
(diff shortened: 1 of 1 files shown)


splitbrain:
What's the status here? Is this mergable or will you want to fix the todos first?

micgro42:
Let me have another look at it this weekend. I want also to check some more templates to ensure that the fields are at least not completely messed up there (c.f. splitbrain/dokuwiki-plugin-bureaucracy#241)

PR #410, opened 2018-08-31
✨ (Import) Add option to create missing pages in csv page schema import

micgro42:
Styling of the checkbox is still a bit off.

WIKI-199

### DIFF
--- a/meta/CSVPageImporter.php
+++ b/meta/CSVPageImporter.php
@@ -7,7 +7,7 @@ class CSVPageImporter extends CSVImporter {
 
     protected $importedPids = array();
 
-    /** @var bool  */
+    /** @var bool[]  */
     protected $createPage = [];
 
     /**


admin/schemas.php
```diff
@@ -196,7 +196,7 @@ protected function html_json(Schema $schema) {
         $form->addFieldsetClose();
 
         $form->addFieldsetOpen($this->getLang('import'));
-        $form->addElement(new \dokuwiki\Form\InputElement('file', 'schemafile'));
+        $form->addElement((new \dokuwiki\Form\InputElement('file', 'schemafile'))->attr('accept', '.json'));
```
splitbrain: does adding method calls on (new foo()) work in 5.6? I remember it didn't work at one time but I forgot when it was introduced. Doesn't addElement() return the new element anyway, so attr could be called on the return of addElement() instead of (new *)?

meta/CSVPageImporter.php
```diff
@@ -7,6 +7,9 @@ class CSVPageImporter extends CSVImporter {
 
     protected $importedPids = array();
 
+    /** @var bool  */
+    protected $createPage = [];
```
splitbrain: What is it? `bool` or an `array`? Maybe `bool[]`?

admin/schemas.php
micgro42: `(new foo())` works in 5.6, but you are right that it is unnecessary. Not sure why I implemented it that way 🤷.
Fixed.

meta/CSVPageImporter.php
micgro42: Started as bool, became an array of bool.
Fixed.

PR #414, opened 2018-09-08
IN operator in struct filter

solewniczak:
This commit introduces additional comparator: "IN" into struct filter. The syntax is:
`structfield IN ("a", "b", "c")`
This is simply translated into:
`structfield = "a" OR structfield = "b" OR structfield = "c"`
I've found the IN syntax more readeable in some situations.

### DIFF
--- a/_test/Search.test.php
+++ b/_test/Search.test.php
@@ -377,4 +377,32 @@ class Search_struct_test extends StructTest {
         $search->addColumn('nope.*');
         $this->assertEquals(0, count($search->getColumns()));
     }
+
+    public function test_filterValueList() {
+        $search = new mock\Search();
+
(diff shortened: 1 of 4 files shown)


splitbrain:
Hmm, you're using the Doku Lexer to parse the filter string into an array, right? That's interesting and quite clever.

I'm not sure about the term `RowValue` - it took me a while to understand what that value is. Would `FilterValueList` make more sense?

solewniczak:
Do you think more tests should be provided?

splitbrain:
Sorry for taking so long. Can you update the documentation?

solewniczak:
Documentation updated. Thank you for merge.

PR #417, opened 2018-09-11
fix the acl check in rpc api

### LINKED ISSUE #344: server error. forbidden to call the method plugin.struct.getData
I'm trying to use struct's xml-rpc api. I have a user that can call the standard wiki rpc functions like wiki.getPage, but when calling a struct function I get a failure:

```
$ cat getData.xml
<?xml version='1.0'?>
<methodCall>
  <methodName>plugin.struct.getData</methodName>
  <params>
    <param>
      <value>
        <string>doku:server:servername</string>
      </value>
    </param>
    <param>
      <value>
        <string></string>
      </value>
    </param>
    <param>
      <value>
        <int>0</int>
      </value>
    </param>
  </params>
</methodCall>
```

Output:

```
$ curl -u <user>:<password> --data @getData.xml -k https://<wiki>/wiki/lib/exe/xmlrpc.php
<?xml version="1.0"?>
<methodResponse>
  <fault>
    <value>
      <struct>
        <member>
          <name>faultCode</name>
          <value><int>-32604</int></value>
        </member>
        <member>
          <name>faultString</name>
          <value><string>server error. forbidden to call the method plugin.struct.getData</string></value>
        </member>
      </struct>
    </value>
  </fault>
</methodResponse>
```

What could be the reason?

elohmeier:
fixes #344

### DIFF
--- a/remote.php
+++ b/remote.php
@@ -41,7 +41,7 @@ class remote_plugin_struct extends DokuWiki_Remote_Plugin {
     public function getData($page, $schema, $time) {
         $page = cleanID($page);
 
-        if(!auth_quickaclcheck($page) < AUTH_READ) {
+        if(auth_quickaclcheck($page) < AUTH_READ) {
             throw new RemoteAccessDeniedException('no permissions to access data of that page');
         }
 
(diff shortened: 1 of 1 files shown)


kasimon:
Thank you!

PR #423, opened 2018-10-08
🐛 Fix autocompletion for tags in lookups

### LINKED ISSUE #422: Tag autocompletion broken for lookup-schemas
The autocompletion of the `Tag` type is broken in lookup-schema, because of a failing check for the pageid.

The autocomplete results are retrieved using the SQL below. This doesn't work for lookups because `D.pid` is just the rownumber. So, these two lines should only be added if we are dealing with a page-schema. However, I don't see an obvious way to access the schema from within the type. 🤔

```php
        if($context->isMulti()) {
            /** @noinspection SqlResolve */
            $sql = "SELECT DISTINCT value
                      FROM multi_{$context->getTable()} AS M, data_{$context->getTable()} AS D
                     WHERE M.pid = D.pid
                       AND M.rev = D.rev
                       AND D.latest = 1
                       AND PAGEEXISTS(D.pid) = 1
                       AND GETACCESSLEVEL(D.pid) > 0
                       AND M.colref = ?
                       AND value LIKE ?
                  ORDER BY value";
            $opt = array($context->getColref(), "%$lookup%");
        } else {
            /** @noinspection SqlResolve */
            $sql = "SELECT DISTINCT col{$context->getColref()} AS value
                      FROM data_{$context->getTable()} AS D
                     WHERE D.latest = 1
                       AND PAGEEXISTS(D.pid) = 1
                       AND GETACCESSLEVEL(D.pid) > 0
                       AND value LIKE ?
                  ORDER BY value";
            $opt = array("%$lookup%");
        }
````

micgro42:
- [ ] refactoring -- currently the `$enforcePageAccess` needs to know about the table being aliased as `D`
- [ ] tests

Fix #422

### DIFF
--- a/types/Tag.php
+++ b/types/Tag.php
@@ -4,6 +4,7 @@ namespace dokuwiki\plugin\struct\types;
 
 use dokuwiki\plugin\struct\meta\QueryBuilder;
 use dokuwiki\plugin\struct\meta\QueryBuilderWhere;
+use dokuwiki\plugin\struct\meta\Schema;
 use dokuwiki\plugin\struct\meta\SearchConfigParameters;
 use dokuwiki\plugin\struct\meta\StructException;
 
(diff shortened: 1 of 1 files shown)


micgro42:
Replaced by #425

Issue #424, opened 2018-10-22
struct header missing on pages created by bureaucracy -- root cause identified!


micgro42:
## The Issue

There exists the stubborn issue that pages created via a bureaucracy template action with struct data don't have the usual struct header rendered when first opened. This issue exists basically since the very beginning of struct, when we added bureaucracy support.
There were several attempts to fix this before, especially by @solewniczak in eaab469c00cfe321e34b8028f8f0004daa52b91e and #355, but the problem still persists and the current code generates PHP Warnings 😕.

## The Cause
I have now finally tracked its root cause:

The bureaucracy template action adds an invisible iframe calling `do=export_html` for every created page:

```php
        foreach ($pages as $ID) {
            // indexerWebBug uses ID and INFO[exists], but the bureaucracy form
            // page always exists, as does the just-saved page, so INFO[exists]
            // is correct in any case
            tpl_indexerWebBug();

            // the iframe will trigger real rendering of the pages to make sure
            // any used plugins are initialized (eg. the do plugin)
            echo '<iframe src="' . wl($ID, array('do' => 'export_html')) . '" width="1" height="1" style="visibility:hidden"></iframe>';
        }
```

See https://github.com/splitbrain/dokuwiki-plugin-bureaucracy/blob/519ab560298fa550e3a2dcd13f3e1dbfc1de163f/helper/actiontemplate.php#L318-L327

However, our struct syntax rendering refuses to render on anything other than the `show` action:
https://github.com/cosmocode/dokuwiki-plugin-struct/blob/8ddbe81a4f5a042937b16b87490bed472f61995b/syntax/output.php#L86-L88

Thus the render's xhtml output is cached without the usual struct header.

## Possible Solutions

I see three possible ways to resolve this:

- have the bureaucracy iframe request the `show` action (not sure if this will break anything)
- have the struct output render on `export_html` as well
- adjust DokuWiki core to not cache the xhtml output for any actions other than `show`

@splitbrain Any ideas which approach would be best to implement?

splitbrain:
I think making struct render in `export_xhtml` makes most sense?

PR #446, opened 2019-04-03
Restore struct info in PDF output

### LINKED ISSUE #300: After upgrade struct data not visible with plugin Dw2pdf
After the last upgrade to Release 2017-02-19b "Frusterick Manners" struct data are not visible any more in pages printed with Dw2pdf. (Formerly struct data were shown on the first page in a small box, which was quite handy for me).

Now I don't know if that is due to struct or the upgrade of the dokuwiki platform but maybe you can check if there's a connection.

Thank you

Johannes

annda:
fixes #300
related to #424 

It adds `export_pdf` to whitelisted actions in struct's render method. Is this the right approach or a band-aid solution (see #424)?

### DIFF
--- a/syntax/output.php
+++ b/syntax/output.php
@@ -100,7 +100,7 @@ class syntax_plugin_struct_output extends DokuWiki_Syntax_Plugin {
         $tables = $assignments->getPageAssignments($ID);
         if(!$tables) return true;
 
-        if($mode == 'xhtml') $R->doc .= self::XHTML_OPEN;
+        $R->doc .= self::XHTML_OPEN;
 
         $hasdata = false;
         foreach($tables as $table) {
@@ -142,10 +142,10 @@ class syntax_plugin_struct_output extends DokuWiki_Syntax_Plugin {
             $R->table_close();
         }
 
-        if($mode == 'xhtml') $R->doc .= self::XHTML_CLOSE;
+        $R->doc .= self::XHTML_CLOSE;
 
         // if no data has been output, remove empty wrapper again
-        if($mode == 'xhtml' && !$hasdata) {
+        if(!$hasdata) {
             $R->doc = substr($R->doc, 0, -1 * strlen(self::XHTML_OPEN . self::XHTML_CLOSE));
         }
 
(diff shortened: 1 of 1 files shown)


splitbrain:
Hmm I think we could make this a little bit more robust (in terms of what people expect) by instead of checking for the current action checking if the renderer is a descendant of `Doku_Renderer_xhtml`.

Or maybe we could use a blacklist instead of a whitelist, since all we don't want to run on is `metadata`, right?

@micgro42, @solewniczak do you have any opinion on this?

micgro42:
It has been some time since I've looked into this, but aren't there some more actions around? I'm thinking of raw text or odt export. Also, I'm not sure I remember how exactly we handled this in the prosemirror plugin?

I think I would favor opening this gradually. Maybe create an explicit `$whitelistedActions` variable? Then we can expand it more easily when requested by other plugins.
On the other hand doing the check on `Doku_Renderer_xhtml` shouldn't break too much either, I hope.

solewniczak:
Checking if the renderer is a descendant of `Doku_Renderer_xhtml ` will be most robust in my opinion.

annda:
I changed the PR to check the renderer type (`Doku_Renderer_xhtml`). I also assumed that all the XTML renderers keep the `xhtml` format/mode, so I removed the mode checks as well.

But now I am no longer sure if it is the best way to go:

Most of the renderer plugins I checked do inherit from `Doku_Renderer_xhtml`. Exceptions like odt or latexport do not work with the present action whitelist either.

It is an improvement over the current state, but perhaps we need another solution?

annda:
Changed whitelist to blacklist after a quick discussion of https://github.com/cosmocode/dokuwiki-plugin-struct/pull/446#issuecomment-483618116 with @splitbrain

solewniczak:
This merge causes that struct data box is shown on actions such as "login", "admin" etc. I think than rather  than using "BLACKLIST_RENDERER" we should use "BLACKLIST_ACT" constant and filter out the actions where struct data box shouldn't be visible.
struct-data-box-on-login-page

splitbrain:
Yeah, I totally forgot about texts rendered during the same request that are unrelated to the main page. I'll open a new PR.

PR #452, opened 2019-05-30
Prefix Summary type with "Abstract" to hide it in Schema Editor.

### LINKED ISSUE #451: Type "Summary" should not be available in schema editor

solewniczak:
Fixes #451

### DIFF
--- a/meta/Column.php
+++ b/meta/Column.php
@@ -176,6 +176,7 @@ class Column {
             $file = basename($file, '.php');
             if(substr($file, 0, 8) == 'Abstract') continue;
             if(substr($file, 0, 5) == 'Trait') continue;
+            if(substr($file, 0, 4) == 'Auto') continue;
             $map[$file] = 'dokuwiki\\plugin\\struct\\types\\' .$file;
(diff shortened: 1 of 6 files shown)


splitbrain:
Hmm, I don't like to prefix a class name with `Abstract` when it's not an abstract class.

solewniczak:
Alternatively we can add an exception in meta/Column.php. Do you think this will be better solution?

splitbrain:
I was thinking of a `isHidden()` method, but that means to instantiate the class before throwing it away again. Maybe using a different prefix than Abstract would be the easiest and still somewhat clean solution?

solewniczak:
Maybe just "Hidden" as a prefix or make isHidden method static.

splitbrain:
Maybe `Auto`, so this becomes `AutoSummary` - it indicates that this is a field that is `auto`matically calculated/filled and can not be used as schema field.

splitbrain:
Looks like some of the tests need adjustments. @solewniczak can you have a look?

PR #453, opened 2019-06-03
Related to #446 Only render output on whitelisted actions

splitbrain:
A check on the renderer class alone turned out to not be sufficient, because xhtml is rendered for wiki internal texts within an ID context sometimes. Eg. for the login intro text.

This keeps the renderer blacklist and only uses the regexp based check when all other checks have been passed.

cc @solewniczak @micgro42


selfthinker:
I've seen this in edit mode, on search results pages, in the admin section, etc.
Using this branch fixes it for me. :+1:

PR #454, opened 2019-06-07
fix: tests adjustments for AutoSummary type

solewniczak:
This commit removes "Summary" column from Column test since it isn't longer visible here.


splitbrain:
thx

PR #461, opened 2019-08-13
Register an additional hook for bureaucracy plugin

annda:


splitbrain:
@annda can you resolve th econflicts? otherwise this is probably fine to merge.

PR #464, opened 2019-09-17
Messages

annda:
The goal is to display field labels instead of field names in error messages and aggregation filters.

### DIFF
--- a/meta/AggregationTable.php
+++ b/meta/AggregationTable.php
@@ -176,7 +176,14 @@ class AggregationTable {
         $fltrs = array();
         foreach($filters as $column => $filter) {
             list($comp, $value) = $filter;
-            $fltrs[] = $column . ' ' . $comp . ' ' . $value;
+
+            // display the filters in a human readable format
+            foreach ($this->columns as $col) {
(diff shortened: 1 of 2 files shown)


meta/AggregationTable.php
```diff
@@ -176,7 +176,14 @@ protected function renderActiveFilters() {
...
             list($comp, $value) = $filter;
-            $fltrs[] = $column . ' ' . $comp . ' ' . $value;
+
+            // display the filters in a human readable format
+            foreach ($this->columns as $col) {
+                if ($column === $col->getFullQualifiedLabel()) {
+                    $column = $col->getTranslatedLabel();
+                }
+            }
+            $fltrs[] = "$column '$value'";
```
splitbrain: Shouldn't the comparator `$comp` still be used here?

annda: I didn't think that `*~` was very user friendly. Can we get rid of it in the filter status message?

annda:
I added the comparators in human readable form to language strings. Is this better?

splitbrain:
LGTM merge it

Issue #418, opened 2018-09-13
Missing fields "hidden" and "hiddenautoinc" for struct data in bureaucracy plugin


Juergen-aus-Koeln:
Hi,

in combination with the bureaucracy plugin there are some missing fields like "hidden" and "hiddenautoinc" to set struct data automatically by certain parameters.

Best regards
Juergen

splitbrain:
Can you elaborate a bit more how you think this should work exactly? Eg. what are you trying to achieve and how should it be implemented in struct?

Juergen-aus-Koeln:
The idea for this issue came up using the template mode of the bureaucracy plugin:

So your questions:
The "hidden" field makes sense with the field "fieldset", e.g.:
```
fieldset "Norm-Art ANSI"               "Norm-Art" "ANSI"
hidden   "Norm-Art-Z00"                "=ANSI" !
fieldset "Norm-Art ASME"               "Norm-Art" "ASME"
hidden   "Norm-Art-Z01"                "=ASME" !
```
And the "hiddenautoinc" is good to define an identifyer for a set of structured data:
```
fieldset "Bearbeitung abschließen"
thanks   "Seite wurde angelegt"
hiddenautoinc "Zaehler_TM" 00000
submit   "Mitteilung anlegen"
```
Because I can't address the parameters in the way that the data plugin does, I have no chance to fill out struct fields. For example "@@Zaehler_TM@@" doesn't work in the struct data input fields of the template referenced by bureaucracy.

The solution could be if the struct data input fields of the template would be able to contain such syntax elements, e.g. 
![grafik](https://user-images.githubusercontent.com/1800472/50981063-6cb9b480-14fa-11e9-938e-d171b268dc20.png)
or much better in combination with other elements (e.g. preceding text or other parameters):
![grafik](https://user-images.githubusercontent.com/1800472/50981207-c28e5c80-14fa-11e9-8552-1ec3cfad4d8f.png)

PR #469, opened 2019-10-23
Add simple autoincrement field type

### LINKED ISSUE #379: Sequential Numbering Field Configuration
This plugin is a great addition to Dokuwiki and I am using it to manage and maintain a controlled document library.  It would be great to have a field configuration that would allow for automatic sequential numbering.  Similar to the way that the numbering plug in works.  Essentially each time a new page is created in a schema controlled namespace it could be issued a sequential number that is unique.  It could require the user to click a button in the inline form to assign the number.

annda:
Prefix "Auto" is reserved so the type is called Increment. It inherits from Decimal but has no config.

Fixes #379 and #418

### DIFF
--- a/helper/field.php
+++ b/helper/field.php
@@ -1,9 +1,11 @@
 <?php
 use dokuwiki\plugin\struct\meta\Column;
 use dokuwiki\plugin\struct\meta\Schema;
+use dokuwiki\plugin\struct\meta\Search;
 use dokuwiki\plugin\struct\meta\StructException;
 use dokuwiki\plugin\struct\meta\Value;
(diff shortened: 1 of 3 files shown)


helper/field.php
```diff
@@ -55,6 +57,11 @@ protected function setVal($value) {
             }
         }
 
+        // handle empty (hidden) autoincrement field
+        if ($this->column->getType() instanceof Increment && ! $value) {
+            $value = $this->column->getType()->getMaxValue() + 1;
+        }
```
splitbrain: I'm not happy with having Type specific code here

PR #470, opened 2019-11-09
Fix FatalError on unknown struct column in bureaucracy forms

solewniczak:
The pull request: #464 causes a fatal error when the "$this->column" is false which may happen when the column is not defined. This commit fixes that.


annda:
Thanks.

PR #472, opened 2019-12-22
fix PHP7.4 deprecation warning

micgro42:
Accessing arrays with curly braces is deprecated in PHP 7.4 and triggers
annoying warnings.


splitbrain:
Thank you.

PR #473, opened 2019-12-25
Add github workflow for phpcs

micgro42:
I actually wanted to also let Travis run on PHP 7.4, but that is broken because `travis.sh` downloads PHPUnit 8 for PHP 7.4 and that is pretty hard incompatible with our current codebase and I'm not sure how to work around that.

So currently this runs phpcs with the Dokuwiki ruleset on the struct code base.

### DIFF
--- a/.github/workflows/phpCS.yml
+++ b/.github/workflows/phpCS.yml
@@ -1,0 +1,30 @@
+name: PHP Code Style
+
+on: [push]
+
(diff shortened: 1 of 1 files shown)


micgro42:
The output that this would produce when merged: https://github.com/micgro42/dokuwiki-plugin-struct/runs/363383008

PR #476, opened 2020-01-16
Enable using page struct data in dw2pdf templates

annda:
This PR makes use of https://github.com/splitbrain/dokuwiki-plugin-dw2pdf/pull/387
It adds the struct data of the page being exported to dw2pdf's replacement array. That way we can use struct placeholders in pdf templates, like `@PLUGIN_STRUCT_schema_fieldname@`

### DIFF
--- a/_test/output.test.php
+++ b/_test/output.test.php
@@ -3,6 +3,7 @@
 namespace dokuwiki\plugin\struct\test;
 
 use dokuwiki\plugin\struct\meta;
+use Doku_Event;
 
 /**
  * @group plugin_struct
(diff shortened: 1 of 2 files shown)


splitbrain:
Could we have a unit test for this?

splitbrain:
Can you add docs to https://www.dokuwiki.org/plugin:struct:integrations please?

Juergen-aus-Koeln:
What is to do if the definition of a schema or a fieldname contains underscore?

annda:
@Juergen-aus-Koeln Good point, I have to fix this.

PR #489, opened 2020-03-23
One schema, different access modes

annda:
The idea is to no longer have different database schemas/tables for pages and lookups. The data should still be accessed differently based on type: page, lookup or serial. For that purpose, all data is stored using the new primary key composed of page id (pid) and row id (rid).

A new type "serial" is introduced. It might be possible to introduce even more types in the future without the need to modify the underlying database structure.

- page: pid = page id, rev = page revision, rid = 0
- lookup: pid = NULL, rev = 0, rid is autoincremented
- serial: pid = page id, rev = 0, rid is autoincremented

Tasks tracked at https://github.com/cosmocode/dokuwiki-plugin-struct/projects/1


helper/serial.php
```diff
@@ -0,0 +1,60 @@
+<?php
+use dokuwiki\plugin\struct\meta\AccessTable;
+use dokuwiki\plugin\struct\meta\AccessTableSerial;
+use dokuwiki\plugin\struct\meta\StructException;
+
+/**
+ * Allows adding a lookup schema as a bureaucracy action
+ *
+ */
+class helper_plugin_struct_serial extends helper_plugin_bureaucracy_action {
```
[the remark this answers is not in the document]
splitbrain: The comment is still wrong, too. If there are any specifics, it could directly inherit from `helper_plugin_struct_lookup`.

action/serial.php
```diff
@@ -0,0 +1,173 @@
...
+use dokuwiki\plugin\struct\meta\SearchConfig;
+use dokuwiki\plugin\struct\meta\StructException;
+use dokuwiki\plugin\struct\meta\Value;
+
+/**
+ * Class action_plugin_struct_serial
+ *
+ * Handle serial editing
+ */
+class action_plugin_struct_serial extends DokuWiki_Action_Plugin
```
splitbrain: This class looks a lot like `action_plugin_struct_lookup`. I guess we could merge them into one or have them both inherit from a common ancestor.

meta/AccessTable.php
```diff
@@ -16,19 +17,22 @@ abstract class AccessTable {
...
 
     /**
-     * Factory Method to access a data or lookup table
+     * Factory method returning the appropriate data accessor (page, lookup or serial)
      *
      * @param Schema $schema schema to load
      * @param string|int $pid Page or row id to access
      * @param int $ts Time at which the data should be read or written, 0 for now
-     * @return AccessTableData|AccessTableLookup
+     * @param int $rid
```
splitbrain: A parameter description would be helpful here.

meta/AccessTableLookup.php
```diff
@@ -88,27 +69,31 @@ public function saveData($data) {
...
         $ok = true;
 
         // insert single values
         $ok = $ok && $this->sqlite->query($singlesql, $opt);
 
-        // get new pid if this is a new insert
-        if($ok && !$this->pid) {
+        // get new rid if this is a new insert
+        if($ok && !$this->rid) {
             $res = $this->sqlite->query('SELECT last_insert_rowid()');
```
splitbrain: Is this still correct? The internal `ROWID` is probably not equal to our `rid`?

meta/Search.php
```diff
@@ -75,14 +77,15 @@ public function addSchema($table, $alias = '') {
...
 
-        if($this->schemas &&
-            (
-                $schema->isLookup() ||
-                reset($this->schemas)->isLookup()
-            )
-        ) {
-            throw new StructException('nolookupmix');
-        }
+        // FIXME is the mixing still relevant?
```
splitbrain: This check probably needs to be more clever, though I am not sure if it's feasible. Whenever multiple schemas are selected, it would need to be ensured that they have PID data. But maybe we can skip this checking and leave it to the user to do sensible aggregations.

meta/Search.php
```diff
@@ -342,7 +358,8 @@ public function getSQL() {
             } else {
                 // first table
 
-                if(!$schema->isLookup()) {
+                // FIXME this breaks page search, a different check is needed
+                if(false) {
```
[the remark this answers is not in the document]
splitbrain: Ohhh. I see what you mean. It's similar to the problem further up at https://github.com/cosmocode/dokuwiki-plugin-struct/pull/489/files#r397089182

I think it should be possible to always add these where clauses but wrap them in parentheses and OR them together with a check on `pid  IS NULL` .

Eg. `WHERE (pid IS NULL OR (GETACCESSLEVEL(pid) > 0 AND PAGEXISTS(pid) =1 AND ....)`

meta/AccessTableLookup.php
annda: Right, this must be replaced with a normal SELECT MAX()

splitbrain: Hmm inserting with `MAX()+1` and later selecting `MAX()` may not be atomic. A wrapping transaction might make it atomic (but I'm not sure what the isolation level for sqlite transactions is). It might be possible to use `SELECT rid FROM <table> WHERE ROWID = last_insert_rowid()`.

action/move.php
```diff
@@ -185,8 +185,8 @@ protected function updateColumnLookup(Schema $schema, Column $col, $old, $new) {
         $tconf = $col->getType()->getConfig();
         $ref = new Schema($tconf['schema']);
         if(!$ref->getId()) return; // this schema does not exist
-        if($ref->isLookup()) return; // a lookup is referenced, nothing to do
-
+        if(!$ref->getTimeStamp()) return; // a lookup is referenced, nothing to do
+        // FIXME does this make sense at all? it's always a lookup, isn't it?
```
splitbrain, line 189: No, counter-intuitively you can have a Lookup-Field pointing to a page schema. Basically it's a lookup of a certain property of pages. The pid will be saved in the lookup field and needs to be updated when the page moves.

action/search.php
```diff
@@ -41,6 +41,7 @@ public function handle_indexing(Doku_Event $event, $param) {
         if(!$tables) return;
 
         foreach($tables as $table) {
+            // FIXME is timestamp relevant here?
```
splitbrain, line 44: It should always be the newest available data be indexed. I believe a timestamp of zero used to load the newest available page data. If that's still the case this should be fine.

meta/AccessTableSerial.php
```diff
@@ -0,0 +1,124 @@
...
+                    $opt[] = null;
+                }
+            } else {
+                $opt[] = $value;
+            }
+        }
+
+        // FIXME select pid too
+        $ridSingle = "(SELECT (COALESCE(MAX(rid), 0 ) + 1) FROM $stable)";
+        $ridMulti = "(SELECT (COALESCE(MAX(rid), 0 ) + 1) FROM $mtable)";
```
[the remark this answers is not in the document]
annda: It turns out that a global row id (not per page) is more straightforward to handle, so let's leave it as it is.

helper/serial.php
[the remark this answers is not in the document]
annda: This helper has been removed.

action/search.php
annda: Hm, this is of course no longer the case, so I changed the lines and now we explicitly pass the current timestamp to the query.

meta/Search.php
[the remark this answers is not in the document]
annda: Maybe my implementation was not clean enough but the suggested solution added unwanted lookups to page aggregations. Because basically it translates into "not page data OR data associated with accessible pages".

I think at this point the new `$idColumn === 'pid'` (join column) is a decent stand-in for `$schema->islookup()` and I'm using it instead in the checks. Unit and manual test look good so far.

PR #502, opened 2020-05-15
Fix migration for structcombolookup plugin

solewniczak:
This simple change fixes the migration for all wikis that use [structcombolookup](https://www.dokuwiki.org/plugin:structcombolookup) plugin.


splitbrain:
I'm merging this as an immediate fix. On the long term we probably need to have a way that plugins like this can run their own migrations.

cc @annda

PR #503, opened 2020-05-19
Migration 17

annda:
An early version of the previous migration did hot handle lookups pointing to pages properly. That might have corrupted some data for the early updaters. This PR fixes the broken values.


Juergen-aus-Koeln:
Is there a way to recover the lost data?

annda:
This is exactly what this migration is supposed to do. If you are on version 17 of the database and still have missing data, please post more details (export your schemas and some sample data in the Schema Editor).

Juergen-aus-Koeln:
Here are some files, and I hope that you can do something with it:
[Problem with lookup.zip](https://github.com/cosmocode/dokuwiki-plugin-struct/files/4651259/Problem.with.lookup.zip)

In "Eingabe Metadaten.png" you can see the values that are lost. The file "cms_meta.struct.json" includes the superordinate definition. The lookups are definied in "cms_modultypen.struct.json" and "fbg_structstatus.struct.json" (by structstatus plugin). 

Further on there is a label called "cms-modulbeschreibung" which is defined using the structcombolookup plugin and which gives a value based on the value of "cms-modultyp".

Best regards
Juergen

annda:
This is paradoxically good news. Your data is definitely not lost to struct, but the plugins extending struct need to be adjusted to handle it in a new way.

Work on structstatus is ongoing, you can track the progress in the issue you reported earlier https://github.com/cosmocode/dokuwiki-plugin-structstatus/issues/10 

The developer of structcombolookup @solewniczak is also generally aware of the problems with the recent upgrade. It will be helpful if you create an issue there directly and link back here for more context, so we can work on solutions together.

Let's move the discussion away from this PR to the respective plugin repos.

Juergen-aus-Koeln:
Thank you for this hints!

One more question:
Is it possible resp. useful to enter new lookup data despite the misbehavior? Or would this data then prevent a recovery of the old data?

Issue #371, opened 2018-01-18
Option to display struct data at end of page


Tuxdiver:
Hello,

I would like to have an option somewhere (maybe in the struct_assignments page) to insert the structured data at the end of the page instead  of inserting it after the first header.

The code to change seems to be action/output.php, but I don't understand the current code and the dokuwiki integration, so I can't send you a PR...

Best regards,

Dirk

Tuxdiver:
OK, after looking again, it's easy....
```
--- output.php_orig	2018-01-18 15:35:23.451029347 +0100
+++ output.php	2018-01-18 15:38:51.290245850 +0100
@@ -56,12 +56,8 @@
             }
         }

-        // insert our own call after the found position
-        array_splice(
-            $event->data->calls,
-            $ins+1,
-            0,
-            array(
+	// append our own call at the end
+	array_push($event->data->calls,
                 array(
                     'plugin',
                     array(
@@ -69,7 +65,6 @@
                     ),
                     $pos
                 )
-            )
         );
     }
```

This puts the data at the end of the page, so our installation is fine at the moment. But having an option to configure this would be cool!

annda:
We have finally added a configuration setting for this.

PR #509, opened 2020-06-08
Option to display page data at the bottom

annda:
Resolves issue #371

### DIFF
--- a/action/output.php
+++ b/action/output.php
@@ -45,37 +45,47 @@ class action_plugin_struct_output extends DokuWiki_Action_Plugin
     {
         global $ID;
         if (!page_exists($ID)) return;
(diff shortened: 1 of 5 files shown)


action/output.php
```diff
@@ -45,37 +45,47 @@ public function handleOutput(Doku_Event $event, $param)
...
+                        array(
+                            'struct_output', array('pos' => $pos), DOKU_LEXER_SPECIAL, ''
+                        ),
+                        $pos
+                    )
                 )
-            )
-        );
+            );
+        }
```
splitbrain, line 88: Instead of the huge if/then thing I would only do the $ins/$pos setup in the if/then and leave the array_splice as is.

PR #513, opened 2020-07-20
Allow deleting all values from a multi field in inline editor

### LINKED ISSUE #439: Can't delete all data in a multiple field of a lookup schema
If you have a field (i.e. type text) with **multiple** values checked, it's not possible to delete all values **inline**. The editor wants at least one character, otherwise you get the old value(s).

annda:
Fixes #439

Puts a mechanism in place to delete any previous values when saving an
empty multi field. This does NOT apply to page data, which is versioned
and should not be deleted on update.

### DIFF
--- a/meta/AccessTable.php
+++ b/meta/AccessTable.php
@@ -210,7 +210,7 @@ abstract class AccessTable
                 // copy first value to the single column
                 if (isset($value[0])) {
                     $this->singleValues[] = $value[0];
-                    if (strlen($value[0]) === 0) {
+                    if ($value[0] === '') {
                         $this->handleEmptyMulti($this->pid, $this->rid, $colrefs[$colname]);
                     }
                 } else {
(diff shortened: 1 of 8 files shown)


meta/AccessTable.php
```diff
@@ -208,6 +210,9 @@ public function saveData($data)
                 // copy first value to the single column
                 if (isset($value[0])) {
                     $this->singleValues[] = $value[0];
+                    if (strlen($value[0]) === 0) {
```
splitbrain, line 213: Why not `$value === ''`?

annda: Absolutely no good reason, fixed.

annda:
I added simple tests for the new behaviors.

Chris75forumname:
Hi, I cannot delete last remaining value from a multi field in inline editor (field type media). 
I can overwrite/delete 3rd entry, 2nd entry and so on, but if I try to overwrite 1st entry with empty input, the dialog quietly dies and leaves the entry in place (without error message).
Is this related?

PR #518, opened 2020-07-29
Bureaucracy integration: use earlier event to handle lookup fields

### LINKED ISSUE #483: Bureaucracy : struct_field or struct_schema with lookup table => Wrong placeholder replacement
Something wrong when trying to create page with action template (from bureaucracy plugin) with struct_field or struct_schema field. 
In the bureaucracy form, we can see the correct labeled value (inquire from lookup page editor) but when we click on submit button, the placeholder for targeted page seems to be replaced by the primary key (integer) instead of labeled value when refer to lookup table (work correctly with dropdown table)

**Dokuwiki** version: 2018-04-22b "Greebo"
**Bureaucracy** plugin version: 2019-10-28
**Sqlite** plugin version: 2019-03-20
**Struct** plugin version: 2020-02-04

Step to reproduce:
-	Create lookup schema (type text)
-	Populate lookup schema from page
-	Create page schema (type lookup which points to previous populated lookup schema)
-	In page, create bureaucracy form with struct_field or struct_schema fields setup with previous page schema (with action template)
-	Click on submit button 

The page name look like 1_2_1_1 insteat of "\<label1>\_\<label2>\_\<label1>\_\<label1>"

annda:
Switching to the earlier event allows replacing values in page ids. It does not affect data storage.

Fixes #483

### DIFF
--- a/action/bureaucracy.php
+++ b/action/bureaucracy.php
@@ -34,7 +34,7 @@ class action_plugin_struct_bureaucracy extends DokuWiki_Action_Plugin
      */
     public function register(Doku_Event_Handler $controller)
     {
-        $controller->register_hook('PLUGIN_BUREAUCRACY_TEMPLATE_SAVE', 'BEFORE', $this, 'handleLookupFields');
+        $controller->register_hook('PLUGIN_BUREAUCRACY_PAGENAME', 'BEFORE', $this, 'handleLookupFields');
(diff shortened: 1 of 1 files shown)


annda:
A test fails because as of now it requires the latest version of bureaucracy, with [this PR](https://github.com/splitbrain/dokuwiki-plugin-bureaucracy/pull/281) merged.

PR #529, opened 2020-11-02
Fix formatting on German intro page

phallobst:


annda:
Thanks for the contribution!

PR #530, opened 2020-11-13
Additional placeholder for use in headerfiles

moz0817:
This placeholder behaves like NS on start pages and like ID on wiki pages.
I use this e.g. as a filter on a global change.log and define the aggretation in a common headerfile. 
On a start page I'll see all changes inside the namespace. On a wiki page only those of the page.
And all of this with one header and two additional lines of code ;-)


iainhallam:
What if the user has changed the start page name? My wikis use `:home` instead of `:start`.

PR #533, opened 2020-11-20
Add option to remove single columns from wildcard select

saerdnaer:
This allows the user to remove specific columns from wildcard lookup, e.g.

```
---- struct table ----
schema: foo
sort: plz
cols: %pageid%, *, -geo, -map
dynfilters: 1
----


```


splitbrain:
This should have tests.

PR #545, opened 2021-01-01
Fix struct editor missing on DokuWiki master

micgro42:
In the big refactoring of html.php in Dokuwiki, the events like `HTML_EDITFORM_OUTPUT` were removed and replaced with `FORM_EDIT_OUTPUT`. See splitbrain/dokuwiki#3198

This PR adds support for the new event and maintains the support for the legacy one. It is expected to work both on the currently released version of DokuWiki and the current tip of the development branch.

This is broken since mid-October. I'm surprised nobody has complained so far?

### DIFF
--- a/action/edit.php
+++ b/action/edit.php
@@ -7,6 +7,7 @@
  * @author  Andreas Gohr, Michael Große <dokuwiki@cosmocode.de>
  */
 
+use dokuwiki\Form\Form;
 use dokuwiki\plugin\struct\meta\AccessTable;
 use dokuwiki\plugin\struct\meta\Assignments;
 use dokuwiki\plugin\struct\meta\Value;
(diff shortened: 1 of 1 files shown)


action/edit.php
```diff
@@ -34,17 +35,49 @@ public function register(Doku_Event_Handler $controller)
...
         $controller->register_hook('HTML_EDITFORM_OUTPUT', 'BEFORE', $this, 'handleEditform');
+        $controller->register_hook('FORM_EDIT_OUTPUT', 'BEFORE', $this, 'addFromData');
+    }
+
+    /**
+     * Adds the html for the struct editors to the edit from
+     *
+     * Handles the FORM_EDIT_OUTPUT event
+     */
+    public function addFromData(Doku_Event $event, $_param): bool
```
annda, line 46: This declaration should be changed to retain compatibility with PHP 5.6, still officially supported by DokuWiki.

action/edit.php
```diff
@@ -34,17 +35,49 @@ public function register(Doku_Event_Handler $controller)
...
+
+        /** @var Doku_Form $form */
+        $form = $event->data;
+        $pos = $form->findElementById('wiki__editbar'); // insert the form before the main buttons
+        $form->insertElement($pos, $html);
+
+        return true;
+    }
+
+    private function getEditorHtml(): string
```
annda, line 80: PHP 5.6 incompatible declaration

annda (changes requested):
Excellent, thank you for the fix. Just please adjust the two lines addressed by comments.

action/edit.php
micgro42: Hogfather _still_ supports PHP 5.6? 🧐

Fixed.

action/edit.php
micgro42: Fixed.

PR #547, opened 2021-01-08
Implement twig templating for page struct data table

fschrempf:
Currently the struct data assigned to a page is shown as a simple table
at the top of the page. To allow easy customization of this view without
messing with the code, we add a twig template parser that looks for a
matching template file at `lib/plugins/struct/templates/<schema>.twig`.

In case no template is available the renderer falls back to the previous
default implementation.


annda:
This PR seems to me to be an overkill. If you have access to the code of your DokuWiki installation, you can already adjust the display to your liking via [userscript.js](https://www.dokuwiki.org/devel:javascript?s[]=userscript).

Also, including a library as dependency would mean additional hassle with maintenance and builds. 

Can you make a convincing case for adopting this?

fschrempf:
> This PR seems to me to be an overkill.

I guess this depends on the point of view.

> If you have access to the code of your DokuWiki installation, you can already adjust the display to your liking via [userscript.js](https://www.dokuwiki.org/devel:javascript?s%5B%5D=userscript).

My use case is not really about small visual changes that can be done by custom CSS or JS. It is rather about being able to define a completely custom view for the struct data per schema. This needs customizing the markup to a degree that is not possible in a clean way by client side manipulation of the DOM.

> Also, including a library as dependency would mean additional hassle with maintenance and builds.

I understand this is not ideal from a maintainer's perspective.

> Can you make a convincing case for adopting this?

I'm not sure it will be convincing, but I can explain what I'm up to.

In my use case the struct data represents information for the wiki page that should be displayed as a header before the actual page content. It is important that the visual representation of the data is adapted to the schema.

Using twig provides an easy and comfortable way to create a fully customized template for each schema that requires it. I think this solution brings a lot of additional value to DokuWiki as it allows to easily combine appealing data visualizations with the power of the wiki features.

To get a better impression of what I'm working at, there's a screenshot at the bottom of this post of a page that contains struct data and uses a twig template to create the data view.

I still have to look if it would be possible to implement this as a separate plugin instead of an addition to the struct plugin. Maybe this would be a better approach.

grafik

splitbrain:
It makes much more sense with the explanation now. But yes, this should be implemented as an additional plugin. Feel free to open a pull request for introducing an event if needed.

PR #548, opened 2021-01-10
Migration cleanup

annda:
Finally removes a temp table left over by migration 16


action/migration.php
```diff
@@ -298,6 +298,28 @@ protected function migration17(helper_plugin_sqlite $sqlite)
...
+    /**
+     * Removes a temp table left over by migration 16
+     *
+     * @param helper_plugin_sqlite $sqlite
+     * @return bool
+     */
+    protected function migration18(helper_plugin_sqlite $sqlite)
+    {
+        $ok = true;
+        $sqlite->query('BEGIN TRANSACTION');
```
splitbrain, line 310: IIRC it should not be needed to create your own transaction. migrations should be wrapped by transactions automatically. In fact I wonder if this shouldn't create an error, because AFAIK sqlite does not support nested transactions.

annda: The PHP migrations are triggered before the `PLUGIN_SQLITE_DATABASE_UPGRADE` event. Only the SQL migrations are wrapped in a transaction. They are executed right after, in a method `\helper_plugin_sqlite::_runupdatefile`, which starts and ends the transaction.

PR #549, opened 2021-01-16
Add config option to disable deleting serial data in aggregations

annda:
This PR adds a simple config option that will render the `delete` button as disabled in rows listing serial data.


splitbrain:
Documentation should mention that this is purely an UI tweak meant to avoid accidental deletion and that this does not provide protection against deletion on API level nor protection against "vandalism".

PR #550, opened 2021-01-24
Add config option to show data on the top of the page

fschrempf:
Currently the default behavior is to show the data below the first heading if it is found, or else on the top of the page.
In addition to the existing option to always show the data on the bottom, let's also add an option to show it always on the top.

To make things simpler we use a single multi option config setting.

### DIFF
--- a/action/output.php
+++ b/action/output.php
@@ -53,7 +53,6 @@ class action_plugin_struct_output extends DokuWiki_Action_Plugin
         if ($this->getConf('bottomoutput')) {
             $ins = count($event->data->calls);
         } else if (!$this->getConf('topoutput')) {
-            $ins = -1;
             foreach ($event->data->calls as $num => $call) {
                 // try to find the first header
                 if ($call[0] == 'header') {
--- a/action/output.php
+++ b/action/output.php
@@ -47,10 +47,12 @@ class action_plugin_struct_output extends DokuWiki_Action_Plugin
         if (!page_exists($ID)) return;
 
         $pos = 0;
+        $ins = -1;
+
         // display struct data at the bottom?
         if ($this->getConf('bottomoutput')) {
             $ins = count($event->data->calls);
-        } else {
+        } else if (!$this->getConf('topoutput')) {
             $ins = -1;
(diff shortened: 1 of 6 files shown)


fschrempf:
Any comments?

annda:
The multi option config makes sense in principle, but the existing setting `bottomoutput` would stop working. It would be better to add another option instead of replacing the old one. Can you update the PR?

fschrempf:
> The multi option config makes sense in principle, but the existing setting `bottomoutput` would stop working. It would be better to add another option instead of replacing the old one. Can you update the PR?

I wonder if there is the possibility of adding some kind of "migration" code to be run when the user updates to a new release of the struct extension and that sets the new `output_position` value based on the old `bottomoutput`.

If that's not possible or too complicated, I can change this PR to keep the old option.

annda:
There is no migration mechanism for configuration and I don't see a suitable event to use for this purpose. Unless you find one, please change the PR.

splitbrain:
Hmm.. I think the proper way to "migrate" here could be to handle it in the plugin's constructor (not sure which class though). the constructor should check if the old setting is set but the new one isn't. In that case internally set the new config accordingly. Not sure how this would best play along the config manager setting though.

fschrempf:
I didn't manage to come up with a proper solution for migrating to the new config option, so I just reworked the PR now to use a separate boolean option beside the existing one.

fschrempf:
@annda Would you mind taking another look at this? Thanks!

action/output.php
```diff
@@ -47,10 +47,12 @@ public function handleOutput(Doku_Event $event, $param)
...
 
         $pos = 0;
+        $ins = -1;
+
         // display struct data at the bottom?
         if ($this->getConf('bottomoutput')) {
             $ins = count($event->data->calls);
-        } else {
+        } else if (!$this->getConf('topoutput')) {
             $ins = -1;
```
annda, line 56: This line is obsolete now, right?

fschrempf: Indeed, thanks! I pushed a fixup for this!

PR #551, opened 2021-01-26
allow serial entries to be included

splitbrain:
It may make sense to define the serial entry mechanism in a central place and simply include it using the include plugin. This patch makes this possible, by always associating serial data with the "main" page instead of the page the syntax is in.


annda:
Looks good. Because I still find all those different `id`s confusing, I tested this manually as well.

PR #558, opened 2021-02-04
Add an event for custom schema data rendering

fschrempf:
In order to hook in a custom renderer for the schema data on the
page we add a new event PLUGIN_STRUCT_RENDER_SCHEMA_DATA and use
the current table representation as default implementation.

### DIFF
--- a/syntax/output.php
+++ b/syntax/output.php
@@ -10,6 +10,7 @@
 use dokuwiki\plugin\struct\meta\AccessTable;
 use dokuwiki\plugin\struct\meta\Assignments;
 use dokuwiki\plugin\struct\meta\StructException;
+use dokuwiki\Extension\Event;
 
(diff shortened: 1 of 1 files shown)


fschrempf:
See here for a use-case description: https://github.com/cosmocode/dokuwiki-plugin-struct/pull/547#issuecomment-770446257

fschrempf:
And here is the first version of a plugin to make use of this: https://github.com/fschrempf/dokuwiki-plugin-structtemplating

fschrempf:
I rebased this to account for the PHP 8 changes in master and fixed one small issue.

@splitbrain @annda As this is a follow-up of #547, would you mind taking a look at this? Thanks!

annda:
Thanks for the contribution.

PR #559, opened 2021-02-17
Change the language-output of the Date-Format 

ooleanderoo:
Converting the Date using strftime in the end.
IMPORTANT: $this->config['format'] refers to the config of the variable in the struct-scheme-editor. The format of the date no longer follows the definition of date_format, it now follows the definition of strftime.


annda:
Closing without merging because it uses two deprecated functions.

PR #571, opened 2021-04-30
Allow to move the struct data input to the top of the edit form

fschrempf:
In order to show the input form at the top rather than below the input for the page content we introduce a new config option.

This change only makes sense in conjunction with splitbrain/dokuwiki#3485 as otherwise the toolbar gets separated from the main wiki content text input.


splitbrain:
We decided we do not want to add this feature

PR #574, opened 2021-05-08
Fix dw2pdf multipage struct output

fschrempf:
When the dw2pdf plugin is used to export multiple pages in a single document, all the pages are rendered in the same context of a single page. In order to include struct data in the PDF, let's skip the id related checks and allow multiple pages rendered in a single instance of syntax_plugin_struct_output.


splitbrain:
The whole mechanism of automatically outputting struct data on pages is somewhat fragile and depending on many variables, thus I'd be hesitant to just merge this. It should at least be accompanied by unit tests, not only for the functionality here but also for all scenarios in which output is automatically added or should not be added (like in includes, sidebars, etc).

Issue #467, opened 2019-10-17
Pagelinks of aggretations are stored in relation reference 


moz0817:
Results of SQL queries (aggretatíon) are entered in the **relation index**, if they contain data of type PAGE. This will automatically register any **wikipage**, containing such queries, as a **backlink** to the contained targets.

I would call that a **mistake**, especially since there are these inconsistencies:
1) A display **limit** only considers the **first query.**
2) These references are saved only if the **page itself is changed**. Subsequent database changes do not update.

P.S. I would be grateful for any hints to turn it off for the first time.  In my wiki, such links are unfortunately very annoying, and I do not want to do without this ingenious plugin. :-)

PR #583, opened 2021-08-27
Fix for Issue #467 (Aggregation and metadata issue)

alexdraconian:
This will fix table aggregation links are stored in metadata. Originally lists and clouds were not rendered metadata, so I assume this line is accidentally excluded in tables.

P.S. It seems SQLite plugin ignores search filter in 'metadata' rendering mode. Because of it, documents with table aggregation may include unnesessary relation data in their metadata. So it's correct that #467 is problem, not intended function.

P.S.2. This code will not 'clear' metadata automatically. Admins must use another plugin for rebuilding metadata if they want to.

### DIFF
--- a/helper.php
+++ b/helper.php
@@ -32,7 +32,6 @@ class helper_plugin_struct extends DokuWiki_Plugin
      * All descendants are also blacklisted.
      */
     const BLACKLIST_RENDERER = [
-        'Doku_Renderer_metadata',
         '\renderer_plugin_qc'
     ];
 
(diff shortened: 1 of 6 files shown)


alexdraconian:
Ok, I found real cause of this problem.

When rendering metadata, global `$INFO` is null. So filter placeholders like `$ID$`, `$PAGE$` cannot be resolved properly(resolved as empty string); This results every pages are returned from SQL query, no matter how filter is set.

So I made changes:

  * If $INFO is null, get page information with dokuwiki global function `pageinfo()`.
  * Re-enabled metadata rendering for table, list, cloud aggregations.

alexdraconian:
I'm closing this pull request since there're still some awakwardness in my code. I will re-open in other pull request when I get time to make fix for it.

alexdraconian:
Replaced with #609.

PR #609, opened 2022-05-15
Fix metadata rendering problem

alexdraconian:
Fix various issues with metadata render.

Current code has following problems, with metadata rendering.
* List and cloud internal links are not be rendered to metadata.
* Table filters ARE NOT be correctly applied to metadata rendering. This means, EVERY pages will be added as internal link to metadata, regardless of table filter options. As a result, backlinks of page will contain non-related pages.
* Dynamic filters and limit parameter ARE applied to metadata rendering. This means every time dynamic filter is applied or user visit another page of table, backlinks will also be changed dynamically. This will delete related pages from backlink.

This pull request resolves above problems with following solution:
* Enable list and cloud metadata rendering
* In metadata rendering, dynamic filters and limit parameter will be ignored. To achieve this, I edited code so that syntax can be rendered twice; metadata and xhtml.

I tested this code with my personal wiki, which has 1,200 pages and a lot of struct aggregation. But if I missed something, please let me know.

### DIFF
--- a/meta/SearchConfig.php
+++ b/meta/SearchConfig.php
@@ -36,10 +36,9 @@ class SearchConfig extends Search
     /**
      * SearchConfig constructor.
      * @param array $config The parsed configuration for this search
-     * @param bool $applyDynamicFilter If this is true, then dynamic filters will not be parsed.
-     *                             This is for metadata rendering.
+     * @param bool $isMetadataRender If this is true, then dynamic filters will not be parsed.
      */
-    public function __construct($config, $applyDynamicFilter=false)
+    public function __construct($config, $isMetadataRender=false)
     {
         parent::__construct();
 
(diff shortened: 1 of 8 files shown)


meta/SearchConfig.php
```diff
@@ -36,8 +36,10 @@ class SearchConfig extends Search
     /**
      * SearchConfig constructor.
      * @param array $config The parsed configuration for this search
+     * @param bool $renderMetadata If this is true, then dynamic filters will not be parsed.
+     *                             This is for metadata rendering.
      */
-    public function __construct($config)
+    public function __construct($config, $renderMetadata=false)
```
splitbrain, line 42: Minor thing: I would rename the parameter to be independent from it's current usage and align it with what it actually does. Eg. `$applyDynamicFilters` or something similar

syntax/cloud.php
```diff
@@ -85,7 +85,6 @@ public function handle($match, $state, $pos, Doku_Handler $handler)
      */
     public function render($mode, Doku_Renderer $renderer, $data)
     {
-        if ($mode != 'xhtml') return false;
```
splitbrain, line 88: This is potentially problematic, the mode might be something we don't understand. Like a XML renderer, or a something else. We can only do this if we never manipulate `$this->doc` ourselves but always call parent renderer methods - I am not sure this is the case?

(this applies to all adjusted render() methods.)

splitbrain (commented):
Thank you for submitting this PR. It looks good already. I would like to see some unit tests for the new behaviour in SearchConfig.

splitbrain:
There's also something wrong in the SearchConfig class (see broken tests). Probably a bracket mismatch somewhere?

syntax/cloud.php
alexdraconian: Alright, then instead of deleting, changing to this maybe resolve potential issue?

``if ($mode != 'xhtml' and $mode != 'metadata') return false;``

alexdraconian:
I only updated ``syntax/output.php`` in commit b25d427, and saw "All checks have failed". I'll find and fix that, but updated code is working normally with my wiki setup.

splitbrain:
Any chance you could add test cases?

alexdraconian:
My apologyze, I didn't write unit test before, so It may take some time.

meta/SearchConfig.php
alexdraconian: Code of __construct not only blocks dynamic filter, but also blocks limit parameter already defined with aggregation syntax. Is ``$applyDynamicFilters'' still valid in this case?

alexdraconian:
I see there're a lot of refactoring of filter algotithm is going on; It seems I have to overhaul everything in here after refactoring is done?

PR #611, opened 2022-07-04
Handle media fields in bureaucracy plugin

annda:
Bureaucracy uses dots in field id, which needs escaping.

### DIFF
--- a/script/functions.js
+++ b/script/functions.js
@@ -67,7 +67,5 @@ function formatString(format) {
  * Custom onSelect handler for struct img button
  */
 window.insertStructMedia = function (edid, mediaid, opts, align) {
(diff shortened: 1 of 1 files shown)


script/functions.js
```diff
@@ -67,5 +67,7 @@ function formatString(format) {
  * Custom onSelect handler for struct img button
  */
 window.insertStructMedia = function (edid, mediaid, opts, align) {
+    // escape dots in field id, bureaucracy plugin uses dots and it can break jQuery selectors
+    edid = edid.split('.').join('\\.');
```
splitbrain, line 71: how about using document.getElementById(edid) instead of building a selector string below?

annda: Thank you, I was blinded by jQuery again.

PR #617, opened 2022-09-13
Eng notation

PhilippCo:
Add engineering notation (kilo, Mega, ..) to values in a struct table


splitbrain:
Thank you. Can you please update the documentation for the decimal type in the wiki?

PR #620, opened 2022-10-09
Translation update (fr)

dokuwiki-translate:
This pull request contains some translation updates.


splitbrain:
cherry-picked this to avoid the merge commits (messup with the translation interface which is hopefully fixed now)

PR #628, opened 2023-02-25
PHP8 fix

asivery:
This fixes an error in PHP8:
```
struct/meta/ConfigParser.php(45)	TypeError: array_pad(): Argument #1 ($array) must be of type array, bool given
```


splitbrain:
Thanks for the PR. I fixed it in a different way.

asivery:
Great, thanks!

PR #631, opened 2023-03-13
Allow column config in serial syntax

annda:
The same columns are used in aggregation and new entry editor

### DIFF
--- a/action/aggregationeditor.php
+++ b/action/aggregationeditor.php
@@ -152,8 +152,8 @@ class action_plugin_struct_aggregationeditor extends DokuWiki_Action_Plugin
         global $INPUT;
         global $lang;
 
-        $data = $INPUT->arr('data');
-        $tablename = $data['schema'];
+        $searchconf = $INPUT->arr('searchconf');
+        $tablename = $searchconf['schemas'][0][0];
 
         $schema = new Schema($tablename);
         if (!$schema->isEditable()) {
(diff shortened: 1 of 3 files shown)


script/AggregationEditor.js
```diff
@@ -140,7 +142,7 @@ const AggregationEditor = function (idx, table) {
         DOKU_BASE + 'lib/exe/ajax.php',
         {
             call: 'plugin_struct_aggregationeditor_new',
-            schema: schema
+            data: {schema: schema, searchconf: searchconf}
```
splitbrain, line 145: Isn't the schema part of the SearchConfig? Wouldn't it be enough to pass the latter param?

annda: SearchConf contains an array of schemas and I'd rather not refactor the code at this point.

action/aggregationeditor.php
```diff
@@ -164,7 +164,18 @@ protected function newRowEditor()
         echo '<legend>' . $this->getLang('lookup new entry') . '</legend>';
         /** @var action_plugin_struct_edit $edit */
         $edit = plugin_load('action', 'struct_edit');
-        foreach ($schema->getColumns(false) as $column) {
+
+        // filter columns based on searchconf cols from syntax
+        $columns = [];
+        if (!empty($INPUT->arr('data')['searchconf']['cols']) && is_array($INPUT->arr('data')['searchconf']['cols'])) {
+            foreach ($INPUT->arr('data')['searchconf']['cols'] as $col) {
+                $columns[] = $schema->findColumn($col);
```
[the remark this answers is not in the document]
annda: I added a filter for empty/false columns.

script/AggregationEditor.js
splitbrain: Hmm. We only support a single schema anyway, right? Can't you simply access searchconf->schemas[0]? Maybe I misunderstand.

PR #605, opened 2022-04-14
Fixed issue #437, filtering DateTime

cmacmackin:
Applying a filter on a DateTime column in an aggregation was causing an uncaught error to occur and prevent the page being rendered at all (see #437). Fortunately, this turned out to be a very simple issue to resolve; it was just an issue with a variable not being declared in the right place.

PR #638, opened 2023-04-05
Schema to test filtering DateTime data

cmacmackin:
Adding the JSON schema data for the unit test added in #605.


cmacmackin:
I have no idea why that one test failed. It appears to be completely unrelated to this change. Also not clear why it would only happen with PHP 8.0.

splitbrain:
Yeah that failure is unrelated. Thank you for the PR.

PR #644, opened 2023-06-22
Nested Lists

splitbrain:
Currently the list aggregation is more of an after thought than a real useful aggregation. This patch starts to improve on that. It adds a new config `nested` which accepts an integer. This integer defines how many of the selected columns should be used to nest the list.

Example. A normal search result might return:

* audi a80 andi
* audi quattro karsten
* bmw i3 detlef
* bmw i3 jörg

Using a nesting level of 1 would use the first column as grouping:

* audi
  * a80 andi
  * quattro karsten
* bmw
  * i3 detlef
  * i3 jörg

A nesting level of 2 would use the first two columns:

* audi
  * a80
    * andi
  * quattro
    * karsten
* bmw
  * i3
    * detlef
    * jörg

When one of the nesting columns uses multivalues, the values are split into single values and the nested subnodes duplicated under each of their parent multivalues.

Output XHTML has been improved to make styling easier.

Still open:

* allow setting a wrapper class to ease styling
* have a way to nest by first letter (undecided on that)

### DIFF
--- a/_test/ConfigParserTest.php
+++ b/_test/ConfigParserTest.php
@@ -82,7 +82,8 @@ class ConfigParserTest extends StructTest
                 ],
             'csv' => true,
             'target' => '',
-            'align' => ['right', 'left', 'center', null]
+            'align' => ['right', 'left', 'center', null],
+            'nesting' => 0,
         ];
 
         $this->assertEquals($expected_config, $actual_config);
--- a/_test/InlineConfigParserTest.php
+++ b/_test/InlineConfigParserTest.php
@@ -51,6 +51,7 @@ class InlineConfigParserTest extends StructTest
             'summarize' => false,
             'target' => '',
             'widths' => [],
+            'nesting' => 0,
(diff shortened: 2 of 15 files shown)


splitbrain:
This now has indexing and custom class features.

@annda I know that you have some struct test setup on your local machine. It would be great if you could give this branch a quick checkover if everything still works. A couple of CSS classes for aggregations changed and this might cause trouble, especially with the JavaScript editors....

annda (approved):
Looks good.

I think two changes would improve the lists:
1. Make sorting optional  (at least on the top level)
2. Introduce a placeholder for empty values. It makes sense to keep them in a list, but empty list items look somehow wrong.

splitbrain:
Placeholder has been added. Not sure how to best implement the (non)search, yet.

splitbrain:
Sorting is now only done for the index level. @annda would be great if you could have another look with your test data.

PR #646, opened 2023-06-28
New syntax component: filter

annda:
This introduces a new component "filter" which accepts the same syntax as "table". It displays a form with available column values as radio buttons. It is built on top of the existing handling of dynamic search filters via `flt` GET parameters.


splitbrain:
There are some visual issues...

* It seems the header syntax is currently ignored
* HTML escaping seems to be wrong (double escaped)
* Wide filter names are missing the drop down element
* the alignment on multiline values isn't ideal - it would be nicer to have the second row indented, probably an easy fix using display:flex
* I would completely remove the fieldset and intro text
* It would be nice if filters would use the full available width (with some min-width and possibly a max-width too), again this might be achievable with flex and grow/shrink

screenshot-localhost-2023 07 03-16_16_21

splitbrain:
The filters are currently radio boxes. This means I can only ever filter for one thing:

screenshot-localhost-2023 07 03-16_29_43

There is no way to filter for items matching "Schlüsselaktivitäten" OR "Kundenbeziehungen". I wonder if you implemented it this way because our dynfilters do not allow for multiple filters on the same field or if it was a deliberate choice on your part?

meta/Filter.php
```diff
@@ -0,0 +1,130 @@
...
+            $form->addTagOpen('div')->addClass('current');
+            $form->addHTML($colData['label']);
+            $form->addTagClose('div');
+
+            $form->addTagOpen('ul')->attr('aria-expanded', 'false');
+
+            $i = 0;
+            foreach ($values as $value) {
+                $form->addTagOpen('li');
+                $form->addRadioButton(SearchConfigParameters::$PARAM_FILTER . "[$qualifiedColName*~]")
```
splitbrain, line 77: This uses a contains comparison, but since we have the exact values in a dropdown, an exact comparison should be better?

meta/Filter.php
```diff
@@ -0,0 +1,130 @@
...
+
+        $this->renderer->doc .= $form->toHTML();
+    }
+
+    /**
+     * Get all values from current search result grouped by column
+     *
+     * @return array
+     */
+    protected function getAllColumnValues(): array
```
splitbrain, line 105: Would it be useful to have a test for this method?

meta/Filter.php
```diff
@@ -0,0 +1,130 @@
...
+        $form->addHTML('<legend>' . $lang['title'] . '</legend>');
+        $form->addHTML('<p>' . $lang['intro'] . '</p>');
+
+        $form->addTagOpen('div')
+            ->addClass('advancedOptions');
+
+        // column dropdowns
+        foreach ($colValues as $colName => $colData) {
+            $qualifiedColName = $colName[0] !== '%' ? "$schema.$colName" : $colName;
+            $values = array_unique($colData['values']);
```
splitbrain, line 60: I think getAllColumnValues() should return unique values already.

meta/Filter.php
```diff
@@ -0,0 +1,130 @@
...
+                // handle multiple values
+                if (is_array($opt)) {
+                    $colValues[$colName]['values'] = array_merge($colValues[$colName]['values'], $opt);
+                } else {
+                    $colValues[$colName]['values'][] = $opt;
+                }
+            }
+        }
+
+        return $colValues;
```
splitbrain, line 128: Besides ensuring uniqueness here (see above) it probably also makes sense to sort the values? Alphabetically? Or maybe be frequency?

script/StructFilter.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Struct Filter
+ *
+ * remove toggling inherited by reusing core search form
+ */
+jQuery('fieldset.struct-filter-form.search-form button.toggleAssistant').remove();
+jQuery('fieldset.struct-filter-form.search-form .advancedOptions').show();
```
splitbrain, line 7: These come from reusing the  default search styles? Can we avoid adding and removing stuff here?

meta/Filter.php
```diff
@@ -0,0 +1,130 @@
+<?php
+
+namespace dokuwiki\plugin\struct\meta;
+
+use dokuwiki\Form\Form;
+
+/**
+ * Struct filter class
+ */
+class Filter
```
splitbrain, line 10: when rebasing on #644 this will need to inherit from `Aggregation`

syntax/filter.php
```diff
@@ -0,0 +1,115 @@
...
+ * @author  Andreas Gohr, Michael Große <dokuwiki@cosmocode.de>
+ */
+
+use dokuwiki\plugin\struct\meta\AggregationCloud;
+use dokuwiki\plugin\struct\meta\ConfigParser;
+use dokuwiki\plugin\struct\meta\Filter;
+use dokuwiki\plugin\struct\meta\SearchConfig;
+use dokuwiki\plugin\struct\meta\StructException;
+
+class syntax_plugin_struct_filter extends DokuWiki_Syntax_Plugin
```
splitbrain, line 16: when rebasing on #644 this probably should inherit from `syntax_plugin_struct_table`  and then `$this->illegalOptions` should be configured accordingly.

splitbrain (commented):
This looks fine already, but could be improved. Please also rebase on #644 so we can merge both PRs together.

splitbrain:
Hmm. I'm confused. Didn't you rebase your branch on top of nestedlists? I changed the base of the PR but still get a bunch of commits that have nothing to do with your changes?

meta/Filter.php
annda: Done.

meta/Filter.php
annda: Done.

meta/Filter.php
annda: Theoretically you are right. But for some reason input is not processed correctly when the name contains `=` and automatic pre-filling does not work. So instead of handling the `checked` status on my own I decided to keep the `like' operand.

annda:
> HTML escaping seems to be wrong (double escaped)

This is a bug in the Form class, I will submit a PR later:

`$this->addLabelHTML(hsc($label), $for, $pos);` which escapes the content again: `$element = new LabelElement(hsc($content));`

annda:
> Wide filter names are missing the drop down element

Fixed.

> the alignment on multiline values isn't ideal - it would be nicer to have the second row indented, probably an easy fix using display:flex

Fixed.

annda:
> I would completely remove the fieldset and intro text

I removed the intro text. But I need the fieldset for the search styles and Javascript to work.

PR #660, opened 2023-08-08
Metadata render fix

alexdraconian:
Replacement of #609, after filter refactoring.

> Fix various issues with metadata render.
>
> Current code has following problems, with metadata rendering.
>
> List and cloud internal links are not be rendered to metadata.
> Table filters ARE NOT be correctly applied to metadata rendering. This means, EVERY pages will be added as internal link to metadata, regardless of table filter options. As a result, backlinks of page will contain non-related pages.
Dynamic filters and limit parameter ARE applied to metadata rendering. This means every time dynamic filter is applied or user visit another page of table, backlinks will also be changed dynamically. This will delete related pages from backlink.
This pull request resolves above problems with following solution:
>
> Enable list and cloud metadata rendering
In metadata rendering, dynamic filters and limit parameter will be ignored. To achieve this, I edited code so that syntax can be rendered twice; metadata and xhtml.
I tested this code with my personal wiki, which has 1,200 pages and a lot of struct aggregation. But if I missed something, please let me know.


splitbrain:
@alexdraconian currently all tests fail can you check what's up with that? Also could you add tests for the functionality you're changing here?

alexdraconian:
Fixed PHP 8.0 errors, but PHP 7.4 error seems not coming from tests here.

Here's what the error displays:

`PHP Fatal error:  Uncaught ParseError: syntax error, unexpected 'include' (T_INCLUDE), expecting identifier (T_STRING) in /home/runner/work/dokuwiki-plugin-struct/dokuwiki-plugin-struct/lib/plugins/include/_test/GeneralTest.php:3`

I think it's coming from include plugin?

https://github.com/dokufreaks/plugin-include/blob/af0799c0415a994f663dbf88ad98bdaf6582ca31/_test/GeneralTest.php#L3C1-L3C40

alexdraconian:
@splitbrain Done fixing errors and adding test.

PR #663, opened 2023-08-12
Fix a whitespace issue in aggregation table

### LINKED ISSUE #661: Whitespace issue in aggregation table with a wiki type column
It is caused by a non-zero margin-bottom of the last block element (and a non-zero margin-top of the first block element in the print version).

ericguirbal:
In a cell with data of type wiki, the non-zero margin-bottom of the last block element (and the non-zero margin-top of the first block element in the print version) creates an unpleasant whitespace.

This fixes #661.

### DIFF
--- a/print.less
+++ b/print.less
@@ -20,3 +20,17 @@
 .dokuwiki .struct_entry_form {
   display: none;
 }
+
+.dokuwiki .structaggregation {
+  &.structaggregationtable, &.structaggregationeditortable {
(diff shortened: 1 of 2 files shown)


splitbrain:
could you provide a before/after screenshot please?

ericguirbal:
Sure.

 In the browser (before to the left, and after to the right) :
web

The print version (before to the left, and after to the right) :
print

PR #672, opened 2023-09-13
Cleanup

splitbrain:
apply #670 and manually cleanup afterwards
    
* fixes some failing tests
* uses sexplode over array_pad
* reformats multiline arrays


saggi-dw:
> uses sexplode over array_pad

This breaks compatibility with versions under "Jack Jackrum", right? I will change this in the wiki.

Issue #593, opened 2022-01-14
Feature request: multiple namespace search in autocomplete


iainhallam:
At the moment, it looks as though the autocomplete configuration for the Page type can take a single namespace to search in. I have people in one namespace, and groups in another, and I'd like to be able to have a link to either in a Struct field. At the moment, I can only do this by having Struct search the whole site.

I'd like to be able to give the namespace option as `:people, :groups`, or some other way of specifying search in multiple namespaces.

annda:
I'm not sure that this is currently possible. Autocompletion uses the core method `_ft_pageLookup()` internally, which seems to accept only one namespace parameter. @splitbrain Am I overlooking something?

splitbrain:
You're right. I missed that the namespace is added as an `@parameter`, I thought it's different.

So I see two ways to implement this:

1. run the ft_pageSearch() call multiple times for each namespace, then merge the results
2. run the ft_pageSearch() call without a namespace, then filter the results

I think 2 is the more performant approach

annda:
The `namespace` option has been migrated to a more flexible `filter` (together with `postfix`).

Filter is a regular expression, written without the delimiter '/' for backwards compatibility. There is no need to migrate any configs manually.

You can write arbitrary regular expressions in the autocomplete filter, e.g. `(namespace1|namespace2:subnamespace|:topnamespace)` for multiple namespaces.

Issue #622, opened 2022-11-07
Restrict depth of autocomplete search for Page type


iainhallam:
I have some pages with their own subpages that shouldn't appear in the output of autocomplete, so I would like to be able to specify that the search has a maximum depth. I can imagine that other situations might warrant a minimum depth - this would allow searching multiple namespaces but not returning the namespace pages themselves for instance, so taking a cue from Unix find, both `autocomplete.maxdepth` and `autocomplete.mindepth` would be helpful configuration.

annda:
The `namespace` option has been migrated to a more flexible `filter` (together with `postfix`).

Filter is a regular expression, written without the delimiter '/' for backwards compatibility. There is no need to migrate any configs manually.

You can now define depth by writing a regular expression in the autocomplete filter, e.g. `^:systems:[^:]+:components:([^:]+:){1,2}[^:]+$` will match two levels inside `components`.

PR #685, opened 2023-11-13
Improve autocomplete search for pages

### LINKED ISSUE #579: Bug? Page autocompletion aborts after hyphen (special character)
Hi!

First of all, Dokuwiki is great! I hope that this issue has not been mentioned, yet, or i missed an information, and it is explained halfway decent.

In our school wiki there are a bunch of devices documented via most recent stable dokuwiki with struct plugin where the pagenames are corresponding to their names. For example in case of PCs this name is also the hostname, e.g. the full path is `anlage:pc:stock_1_inf1:inf1-pc-01`.

According to https://www.dokuwiki.org/pagename:

```
Allowed characters are letters, digits and, within names 1), the “special characters” ., - and _."

1) i.e. you cannot start or end names with them, resp. they will be eliminated at start or end of names
```

When I enter a hyphen (`-`) in a lookup table inside type page text entry then one more character is accepted by autocompletion but not filtered correctly. After the second character the autocompletion list aborts and shows no result anymore. This happens to all "allowed special characters", thus `.`, `-` and `_`. 

Works:
struct_01

Aborts:
struct_02

My schema for this field:
struct_03

Anyways, thank you for your great work!

annda:
Fixes #579 and #593

### DIFF
--- a/_test/types/PageTest.php
+++ b/_test/types/PageTest.php
@@ -205,7 +205,7 @@ class PageTest extends StructTest
                 'autocomplete' => [
                     'mininput' => 2,
                     'maxresult' => 5,
-                    'namespace' => '',
+                    'filter' => '',
                     'postfix' => '',
                 ],
             ]
@@ -238,6 +238,10 @@ class PageTest extends StructTest
         ], $page->handleAjax());
     }
 
+    /**
+     * Test deprecated option namespace
+     * @return void
+     */
     public function test_ajax_namespace()
     {
         global $INPUT;
@@ -257,7 +261,7 @@ class PageTest extends StructTest
         $this->assertEquals([['label' => 'syntax (wiki)', 'value' => 'wiki:syntax']], $page->handleAjax());
     }
 
-    public function test_ajax_namespace_multiple()
+    public function test_ajax_filter_multiple()
     {
         global $INPUT;
 
@@ -266,7 +270,7 @@ class PageTest extends StructTest
                 'autocomplete' => [
                     'mininput' => 2,
                     'maxresult' => 5,
-                    'namespace' => '(wiki|foo)',
+                    'filter' => '(wiki|foo)',
                     'postfix' => '',
                 ],
             ]
@@ -279,6 +283,10 @@ class PageTest extends StructTest
         ], $page->handleAjax());
     }
 
+    /**
+     * Test deprecated option postfix
+     * @return void
+     */
     public function test_ajax_postfix()
     {
         global $INPUT;
--- a/_test/types/PageTest.php
+++ b/_test/types/PageTest.php
@@ -10,7 +10,7 @@ use dokuwiki\plugin\struct\types\Page;
 /**
  * Testing the Page Type
  *
- * @group plugin_structp
+ * @group plugin_struct
  * @group plugins
  */
 class PageTest extends StructTest
@@ -21,13 +21,14 @@ class PageTest extends StructTest
         parent::setUp();
 
         saveWikiText('syntax', 'dummy', 'test');
+        saveWikiText('foo:syntax:test_special.characters', 'dummy text', 'dummy summary');
 
         // make sure the search index is initialized
         idx_addPage('wiki:syntax');
         idx_addPage('syntax');
         idx_addPage('wiki:welcome');
         idx_addPage('wiki:dokuwiki');
-
+        idx_addPage('foo:syntax:test_special.characters');
     }
 
     public function test_sort()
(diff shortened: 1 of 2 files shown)


annda:
This would be a good place to address #622. 

TODO: think about consequences of depth for relative namespaces and write appropriate tests

types/Page.php
```diff
@@ -224,4 +220,35 @@ public function filter(QueryBuilderWhere $add, $tablealias, $colname, $comp, $va
...
+            $namespace = '^' . $namespace;
+        }
+        // non-regex namespace?
+        if (PhpString::substr($namespace, 0, 1) !== '/') {
+            $namespace = '(?::|^)' . $namespace ;
+            $namespace = '/' . $namespace . '/';
+        }
+        preg_match($namespace, $searchNS, $matches);
+
+        return !empty($matches);
```
splitbrain, line 252: No need to capture the matches, you can simply `return (bool) preg_match(...)`

_test/types/PageTest.php
```diff
@@ -249,6 +257,28 @@ public function test_ajax_namespace()
...
+    public function test_ajax_namespace_multiple()
+    {
+        global $INPUT;
+
+        $page = new Page(
+            [
+                'autocomplete' => [
+                    'mininput' => 2,
+                    'maxresult' => 5,
+                    'namespace' => '(wiki|foo)',
```
splitbrain, line 269: Here a regexp is configured without any delimiters. Later on in nsMatch() you differentiate between regexes and non regexes by the use of delimiters?

types/Page.php
```diff
@@ -224,4 +220,35 @@ public function filter(QueryBuilderWhere $add, $tablealias, $colname, $comp, $va
...
+
+        // prepare any namespace for preg_match()
+        $searchNS = ':' . $searchNS . ':';
+        // absolute namespace?
+        if (PhpString::substr($namespace, 0, 1) === ':') {
+            $namespace = '^' . $namespace;
+        }
+        // non-regex namespace?
+        if (PhpString::substr($namespace, 0, 1) !== '/') {
+            $namespace = '(?::|^)' . $namespace ;
```
splitbrain, line 247: if the given namespace is supposed to not be a regex, then it should be escaped using `preg_quote_cb()`

types/Page.php
```diff
@@ -224,4 +220,35 @@ public function filter(QueryBuilderWhere $add, $tablealias, $colname, $comp, $va
...
+            $namespace = '/' . $namespace . '/';
+        }
+        preg_match($namespace, $searchNS, $matches);
```
splitbrain, line 250: I guess invalid regexes should be handled somehow? Maybe at least send a `msg()` or probably better throw an exception?

splitbrain (commented):
Currently the regex only is applied to the namespace, not the whole ID. I guess that makes sense from the name of the configuration (`namespace`). More useful IMHO would be a match against the whole ID? Maybe we should think about how we could rename the config without losing existing configuration.

I am also not convinced that we need to differentiate between regex and non-regex configs. Any valid previous configuration should still be a valid regex, so why not always treat the config as regex and add the delimiters ourselves?

types/Page.php
annda: Right.

_test/types/PageTest.php
annda: You are right, this is wrong with the current code.

But I followed your suggestion and got rid of the distinction regex / non-regex, so this definition is now correct.

types/Page.php
annda: All namespaces are now assumed to be regex.

splitbrain:
Hmm maybe overwrite mergeConfig() for the Page type?

types/Page.php
annda: Done.

annda:
I converted `namespace` and `postfix` to `filter`. Can you take another look?

types/Page.php
```diff
@@ -78,25 +100,26 @@ public function handleAjax()
...
-        // this basically duplicates what we do in ajax_qsearch()
+        $filter = $this->config['autocomplete']['filter'];
+
+        // try to use deprecated options namespace and postfix as filter
+        $namespace = $this->config['autocomplete']['namespace'] ?? '';
+        $postfix = $this->config['autocomplete']['postfix'] ?? '';
+        if (!$filter) {
+            $filter = $namespace ? $namespace . ':' : '';
+            $filter .= $postfix ? '.+?' . $postfix . '$' : '';
+        }
```
splitbrain, line 114: Is repeating this here necessary? Isn't `getConfig()` executed when the object is instantiated anyway?

splitbrain (approved):
LGTM

types/Page.php
annda: Repeating the code was necessary because I applied the migration in the wrong place to begin with. I moved it from `getConfig()`  to `mergeConfig()`, which is called when instantiating the object.

The redundant code is removed.

Issue #598, opened 2022-03-21
Feature request: Allowing more flexible JOINs over shemas


cmacmackin:
As far as I can tell this is not currently possible, but it would be extremely useful to me. One of the struct datatypes is a Page, corresponding to a page of the wiki. In aggregations it would be nice to be able to work with data from schemas associated with the pages referenced in each row of the table. I imagine this using syntax inspired by [value aggregations](https://www.dokuwiki.org/plugin:struct:aggregation#value). My use-case for this would be filtering but possibly it could be used for joining as well (unfortunately the [structjoin](https://www.dokuwiki.org/plugin:structjoin) plugin doesn't seem to support joining based on the Page type).

To provide a motivating example, I'm running a wiki for an organisation which holds lots of meetings. Meetings often generate tasks which someone is meant to complete. I've already created schemas for meetings and tasks (each associated to a separate namespace), where tasks have a field called `parent` which refers to a page with which the task is associated. This page is often a meeting but doesn't have to be. On the page documenting a particular meeting I've written aggregations which will display all tasks for which that meeting is the `parent`. This works very well.

We hold some meetings which are public and at which motions may be passed. I've created a schema for motions as well, with a `meeting` field corresponding to the Page of the meeting where the motion was tabled. In many cases it is more appropriate to associate a task with a motion than with the meeting. However, it would still be useful to know about the tasks arising due to the motions passed at a meeting. Therefore, I would like to put in an aggregation which can filter based on tasks for which the `parent` page is a motion whose `meeting` is the current one.

In this particular case there are workarounds (e.g., giving a task multiple parents). However, I feel like such a feature would have many other uses. For example, if date/time filtering can be made to work (#437), it could allow aggregations of all tasks created at meetings during a certain time period.

I'd be happy to contribute a PR to implement this, but as I'm new to PHP and inexperienced with databases I thought I'd raise it here first. Perhaps someone has a suggestion on how to go about implementing it.

cmacmackin:
Thinking about this a bit more, what I really want is a more flexible way to join schemas. Currently in aggregations you can only perform inner joins using the page. That's a perfectly sensible default, but I wonder if there's a way to extend it. I note that the [the `cargo` plugin for MediaWiki](https://www.mediawiki.org/wiki/Extension:Cargo/Querying_data##cargo_query) allows doing arbitrary LEFT OUTER JOINs. Maybe adding a new option along the lines of 
```
[...]

Issue #285, opened 2017-04-23
Join schemas on different pages


iainhallam:
I'm still trying to set up a user information system using DokuWiki, and it looks as though struct will be the best way to achieve this, but I either need struct to implement allowed viewers for certain schemas or to allow joining data from different pages. It seems to me that adding the management overhead of a second permission system in an external database to this complex setup will be problematic, and still needs code changes, so I suspect it will be more useful to myself and others to be able to use the built-in DokuWiki ACLs to control access to this data.

My current setup is to have a namespace, `people:` which contains pages for individuals, e.g., `people:iain.hallam`. I've got a second namespace set up with appropriate ACLs, `people:confidential:` (with pages, e.g., `people:confidential:iain.hallam`) so that information stored there can only be seen by users in a group `confidential`.

I've defined two schemas, `person` and `person_confidential`, and assigned them to the appropriate paths, and I can now edit information on both the person page and the confidential page. My problem comes when I want to create an aggregation page (also restricted to the group `confidential`) that shows public and private information together.

Since the schemas are stored as tables, I'd like to be able to have struct issue a query to join them both:

    SELECT * FROM person LEFT JOIN person_confidential ON person.%page% = person_confidential.%page%;

I guess in struct syntax, this would modify the schemas line of the table output syntax somehow? Perhaps

    ---- struct table ----
    schemas: person < person_confidential
    join: person.%page% = person_confidential.%page%
    cols: *
    ----

Is this a potential approach or is there a better way of protecting only a portion of struct data for an individual page?

juliusverrel:
This would be extremely useful also in the following use case (with multi-valued vield)

  * schema Event with field Projects (multi-valued)
  * schema Project with field Users (mult-valued)
  * left join (Event, Project) would have a field users (or Project.users?), listing for each event all users of any of the projects assigned to this event
  * for instance, this would allow showing each user a list of events related to any of their projects with the following code:

```
---- struct table ----
schemas: Event < Project
join: Event.Projects ~ Project.ID
cols: Event.*
filter: Project.users ~ CURRENTUSER
----
```

PR #691, opened 2023-11-19
Add custom joins

### LINKED ISSUE #269: join two schemas
First of all: thanks for providing these great plugins to the community. I am using both, the struct & bureaucracy plugin, they are great and really appreciated.

What I am looking for is a function to "join" the structured data from different schemas in a simple 1:N manner and it seems with the "lookup dropdown/lookup schema" features it was somehow intended. Maybe I am missing something more general here?

Schema "City":  name, inhabitants, country
Page: Lyon, 500.000, France
Page: Paris, 2.200.000, France
Page: Washington, 650.000, US

Here a quick example:
Schema "Country":  name, president
Page: US, Trump
Page: France, Hollande

Overview/Aggregation page for all Cities:

---- struct table ----
schema: City, Country
cols: City.name, City.inhabitants, Country.name, Country.president
filter: City.country=Country.name

I'd love to have such a feature. Cheers.

cmacmackin:
This introduces the ability to JOIN schemas in aggregations ON fields other than the page ID. It required some significant refactoring of the SQL generation so will likely need extensive review. 

**Currently this PR is a draft. I have begun writing it so I can keep track of all the changes I have made.**

Fixes #269, #285, #598

This has involved the following changes:

- Introducing additional config syntax to specify what fields to join schemas on (plus associated code to process and store this data)
- Setting the "context" for all of the fake columns used by `Search`.
- Moving logic for handling multi-valued fields to within the `select()` method
- Moving datatype-specific select logic out of `select()` and into `selectCol()` (which gets called from `select()`)
- Adding a `getSqlCompareValue()` method to the datatype classes. This returns an expression for the contents of the column to be used either in filtering or joining. If need be, it will also insert any other conditional expressions needed into a query's WHERE clause. 
- Adding a `getSqlConstantValue()` method to the datatype classes. This will wrap a value that is being compared against in any additional logic that is needed (e.g., making it lower-case or converting it to digits).
- Implement  `filter()` only on `AbstractBaseType` and refactor it to use `getSqlCompareValue()` for the type-specific logic.
- Introducing a `joinCondition()` method to construct alternative JOIN ON clauses when generating SQL queries.
- Adding logic to a number of existing methods to handle custom join conditions.
- Updates to existing tests plus writing new tests for the new features.

Tasks still to do:
- [ ] Decide how to handle comparisons with page IDs and/or names in custom joins
- [ ] Decide how (if at all) to handle joins on columns that are multi-valued
- [ ] Write some additional tests to

### DIFF
--- a/_test/AccessTableDataReplacementTest.php
+++ b/_test/AccessTableDataReplacementTest.php
@@ -184,7 +184,7 @@ class AccessTableDataReplacementTest extends StructTest
                         AND PAGEEXISTS(data_bar.pid) = 1
                         AND (
                             data_bar.rid != 0
-                            OR (data_bar_ASSIGNED = 1 OR data_bar_ASSIGNED IS NULL)
+                            OR (ASSIGNED = 1 OR ASSIGNED IS NULL)
                         )
                     )
                 )
--- a/_test/AccessTableDataReplacementTest.php
+++ b/_test/AccessTableDataReplacementTest.php
@@ -184,7 +184,7 @@ class AccessTableDataReplacementTest extends StructTest
                         AND PAGEEXISTS(data_bar.pid) = 1
                         AND (
                             data_bar.rid != 0
-                            OR (ASSIGNED = 1 OR ASSIGNED IS NULL)
+                            OR (data_bar_ASSIGNED = 1 OR data_bar_ASSIGNED IS NULL)
                         )
                     )
                 )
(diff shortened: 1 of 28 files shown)


splitbrain:
@cmacmackin awesome! It will probably take us a while to review this but I appreciate the effort.

cmacmackin:
@splitbrain I've now finished my work on this. I appreciate that this is a very complicated pull request that will take some time to review. I'm not really in a rush. This was just something I started working on months ago and figured I should try to get it finished before tackling the other issues I recently raised.

I've written tests for new features that I've added but I can't speak for how well the existing test-suite covers some of my refactoring.

cmacmackin:
Also, I'm not experienced working with databases, so it is possible that some of the queries could be structured better.

PR #695, opened 2023-12-02
Add option to restrict users/groups that can autocomplete usernames

### LINKED ISSUE #690: Autocomplete security for User datatype
I'm administering a wiki where we would like to keep the names of users confidential (unless the users choose to edit pages). For the most part this works fine, but there is a vulnerability where people can use autocomplete on `User` data to find names of users. I had been thinking of creating a subclass of that type where autocomplete only worked for certain trusted groups of users, but I noticed this note in the `struct` source code:
```
@todo should we have any security mechanism? Currently everybody can look up users
```
If I created a pull request adding a config to restrict user-lookup to certain users or groups, would this be of interest?

cmacmackin:
A configuration "allow_username_autocomplete" has been added. This is a list of usernames and/or groups to which to limit the availability of autocomplete for username type data. By default this is set to @ ALL (space included in this message so I don't send notifications to every repository contributor), so the behaviour is backwards-compatible. Tests of this new feature have been added. I've only added language data for English, as that is the only language I speak. Contributions for other languages would be more than welcome.

Fixes #690

### DIFF
--- a/types/AbstractBaseType.php
+++ b/types/AbstractBaseType.php
@@ -546,6 +546,18 @@ abstract class AbstractBaseType
     }
 
     /**
+     * Convenience method to access plugin configurations
+     *
+     * @param string $string
+     * @return string
+     */
+    public function getConf($string)
+    {
+        if (is_null($this->hlp)) $this->hlp = plugin_load('helper', 'struct');
+        return $this->hlp->getConf($string);
+    }
+
+    /**
      * With what comparator should dynamic filters filter this type?
      *
      * This default does a LIKE operation
--- a/types/User.php
+++ b/types/User.php
@@ -65,7 +65,6 @@ class User extends AbstractMultiBaseType
     {
         /** @var AuthPlugin $auth */
         global $auth;
-        global $conf;
         global $INPUT;
         global $_SERVER;
         global $USERINFO;
@@ -76,7 +75,7 @@ class User extends AbstractMultiBaseType
 
         if (
             !auth_isMember(
-                $conf['plugin']['struct']['allow_username_autocomplete'],
+                $this->getConf('allow_username_autocomplete'),
                 $_SERVER['REMOTE_USER'],
                 (array) $USERINFO['grps']
             )


annda:
Thanks for the PR. 

However, this implementation may not work if the plugin config is not loaded at the time the check takes place. You cannot detect it in your tests, because you manually set the value in the config array. So in a way you simulate a partial config load, which may not yet have happened in a normal request.

I think you can avoid this problem by introducing a wrapper `getConf()` method in the AbstractBaseType, similar to the existing convenience method `getLang()`. This would internally load the config if needed.

cmacmackin:
Thanks for the feedback. My ignorance of the internal workings of DokuWiki is showing. I've added a `getConf()` method as you suggest. Is there a better way to write my tests so that they'd be able to catch problems such as these?

cmacmackin:
The test failures here all seem to have arisen due to recent changes to master.

annda:
Thanks for the changes, the code is more robust now.

On second thought, and seeing how cumbersome it is to handle the config in tests, is there a reason why you want to introduce a plugin setting? It would certainly be much easier and also more flexible to handle this in autocomplete configuration.

cmacmackin:
> On second thought, and seeing how cumbersome it is to handle the config in tests, is there a reason why you want to introduce a plugin setting? It would certainly be much easier and also more flexible to handle this in autocomplete configuration.

I'm not sure that really makes sense from a security perspective. If an admin doesn't want some users being able to see a list of usernames in one schema, surely that would apply to all schemas. Requiring them to configure that for every User column in every schema creates the opportunity for them to forget one.

splitbrain:
I think I would want to set this based on context. A schema used by the marketing department would probably only want to have marketing people in the autocomplete. A schema for bug reports might want to use all the developers in one field (assignee) but all users in another (reporter)...

cmacmackin:
>I think I would want to set this based on context. A schema used by the marketing department would probably only want to have marketing people in the autocomplete. A schema for bug reports might want to use all the developers in one field (assignee) but all users in another (reporter)...

That's a different thing (though obviously useful). That is about who shows up in the autocomplete list. This PR is about who is allowed to see the autocomplete list.

PR #701, opened 2024-01-17
bureucracy: handle multi fields of type page with setting 'usetitles'

### LINKED ISSUE #700: [Syntax Error] Using struct-fields in bureaucracy linking to pages
Yesterday my bureaucracy form worked well. Today I updated the struct plugin. After plugin update I got an error when using the form:

image

The form uses: struct_field "events.otherpartners" !

Definition in struct editor:
image

The selection of the page (autocompletion) works, but after submitting the form, I got the error and "Array" in my form field.

annda:
Fixes #700

### DIFF
--- a/helper/field.php
+++ b/helper/field.php
@@ -158,14 +158,19 @@ class helper_plugin_struct_field extends helper_plugin_bureaucracy_field
      */
     protected function createValue()
     {
-        $preparedValue = $this->opt['value'] ?? '';
+        // input value or appropriately initialized empty value
+        $preparedValue = $this->opt['value'] ?? ($this->column->isMulti() ? [] : '');
 
         // page fields might need to be JSON encoded depending on usetitles config
         if (
             $this->column->getType() instanceof Page
             && $this->column->getType()->getConfig()['usetitles']
         ) {
-            $preparedValue = json_encode([$preparedValue, null], JSON_THROW_ON_ERROR);
+            if ($this->column->isMulti()) {
+                $preparedValue = array_map(static fn($val) => json_encode([$val, null], JSON_THROW_ON_ERROR), $preparedValue);
+            } else {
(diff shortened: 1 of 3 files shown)


moz0817:
Hi annda,
there is a littĺe mistake in helper/field.php Line 172. 
The 2. parameter of array_map should be an array. I test [ $preparedValue, null ] an d it works.

The Failure occurs, when an placeholder **FORMPAGE_*** is used for a **page**-field with **multivalue** and **usetitles** active.

moz0817:
The error is still there, but only by using the placeholder FORMPAGE_ID.

In my opinion, the check for the 'page' type to see if 'usetitles' is enabled—which leads to the error—can be removed here. Wiki IDs are always used for input, after all; the 'usetitles' option essentially applies only to the output.
Here the cleaned section in the function createValue() in helper/field.php


        // page fields might need to be JSON encoded (the distinction regarding usetitles can be omitted here)
        if ( $this->column->getType() instanceof Page ) {
            $preparedValue = json_encode([$preparedValue, null], JSON_THROW_ON_ERROR);
        }
        $value = new Value($this->column, $preparedValue);

helper/field.php
moz0817: The error is still there, but only when using the placeholder FORMPAGE_ID.

In my opinion, the check for the 'page' type to see if 'usetitles' is enabled—which leads to the error—can be removed here. Wiki IDs are always used for input, after all; the 'usetitles' option essentially applies only to the output.
Here the cleaned section in the function createValue() in helper/field.php


        // page fields might need to be JSON encoded (the distinction regarding usetitles can be omitted here)
        if ( $this->column->getType() instanceof Page ) {
            $preparedValue = json_encode([$preparedValue, null], JSON_THROW_ON_ERROR);
        }
        $value = new Value($this->column, $preparedValue);

PR #705, opened 2024-01-29
Lazy search

annda:
Search is currently executed in aggregation constructor. Which means even on initial metadata run (also happening in edit mode), and then again before rendering.

This PR always gets search results via a new method `getRows()`. The search is only executed once and on subsequent calls the existing results are returned.

Also, the `render()` in aggregations reuses a renderer blacklist from the helper, which prevents rendering in metadata.


annda:
@splitbrain I made a few suggested changes, can you please take a quick look again?

meta/SearchResult.php
```diff
@@ -0,0 +1,171 @@
...
+    protected $count = -1;
+
+    /**
+     * Construct SearchResult
+     *
+     * @param \PDOStatemen $res
+     * @param int $rangeBegin
+     * @param int $rangeEnd
+     * @param array $columns
+     * @param bool $pageidAndRevOnly
```
splitbrain, line 29: can we have some parameter descriptions? especially the last one is a bit mysterious ;-)

meta/SearchResult.php
```diff
@@ -0,0 +1,171 @@
...
+                continue;
+            }
+
+            $this->addPid($row['PID']);
+            $this->addRid($row['rid']);
+            $this->addRev($row['rev']);
+            $this->addRow($resrow);
+        }
+
+        $res->closeCursor();
```
splitbrain, line 64: I wonder if the cursor should be closed here or by the one who passes it... If this method does it, we should mention it in the method comment

meta/SearchResult.php
```diff
@@ -0,0 +1,171 @@
...
+            // skip empty rows
+            if ($isempty && !$pageidAndRevOnly) {
+                $this->decreaseCount();
+                continue;
+            }
+
+            $this->addPid($row['PID']);
+            $this->addRid($row['rid']);
+            $this->addRev($row['rev']);
+            $this->addRow($resrow);
```
splitbrain, line 61: These not longer need to be functions but can be set directly.

meta/SearchResult.php
```diff
@@ -0,0 +1,171 @@
...
+    public function getRevs(): array
+    {
+        return $this->revs;
+    }
+
+    /**
+     * @param string $pid
+     * @return void
+     */
+    public function addPid($pid)
```
splitbrain, line 112: these setters can now be dropped

PR #716, opened 2024-05-13
Remove leading zeros in Decimal type.

gregoryloichot:
Fix a bug. It was possible to create Decimal like 00123. This Decimal could then be edited to 100123, but could not be edited back to 00123. This pull request remove the leading zeros.


types/Decimal.php
```diff
@@ -99,7 +99,7 @@ public function renderValue($value, \Doku_Renderer $R, $mode)
     public function validate($rawvalue)
     {
         $rawvalue = parent::validate($rawvalue);
-        $rawvalue = str_replace(',', '.', $rawvalue); // we accept both
+        $rawvalue = trim((float)str_replace(',', '.', $rawvalue), 0); // we accept both
```
splitbrain, line 102: If my value is `1000` this will reduce it to `1`.

```
echo trim((float) '1000', 0);
1
```

 I believe casting the value to float should already do what you want. 

```
echo (float) '00004440000';
4440000
```

types/Decimal.php
```diff
@@ -113,7 +113,9 @@ public function validate($rawvalue)
             throw new ValidationException('Decimal max', (float) $this->config['max']);
         }
 
+        // remove leading zeros
```
splitbrain, line 116: this comment seems to be in the wrong place

types/Decimal.php
```diff
@@ -99,7 +99,7 @@ public function renderValue($value, \Doku_Renderer $R, $mode)
     public function validate($rawvalue)
     {
         $rawvalue = parent::validate($rawvalue);
-        $rawvalue = str_replace(',', '.', $rawvalue); // we accept both
+        $rawvalue = trim((float)str_replace(',', '.', $rawvalue), 0); // we accept both
 
         if ((string)$rawvalue != (string)(float) $rawvalue) {
             throw new ValidationException('Decimal needed');
```
splitbrain, line 105: I am wondering. Shouldn't this already prevent the issue you're trying to fix?

splitbrain (changes requested):

types/Decimal.php
gregoryloichot: I think you are right. I will change that.

types/Decimal.php
gregoryloichot: Thanks for your comments.

I updated the code. Let me know if you have any comments.

PR #732, opened 2025-03-08
Add ability to get global structure data by remote API

MaksOn-dev:
To get global data by using `plugin.struct.getData` method leave `page` parameter (first) empty and specify `schema` (second) one.


springfielddatarecovery:
We paid this guy on upwork to make this patch. Tested today and it works. Would love to see it merged into the official codebase.

